PR #1, opened 2020-10-08
Added type check to Get(string id).

RichMercer:
Ensures the item returned is of the same type as requested.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
+++ b/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
@@ -46,13 +46,11 @@ namespace Microsoft.Azure.CosmosRepository
 
             TryLogDebugDetails(_logger, () => $"Read: {JsonConvert.SerializeObject(item)}");
 
-            if (item.Type == typeof(TItem).Name)
-                return item;
-
-            return null;
+            return string.IsNullOrEmpty(item.Type) || item.Type == typeof(TItem).Name ? item : null;
         }
 
         /// <inheritdoc/>
+        /// <remarks></remarks>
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -46,7 +46,10 @@ public async ValueTask<TItem> GetAsync(string id)
 
             TryLogDebugDetails(_logger, () => $"Read: {JsonConvert.SerializeObject(item)}");
 
-            return item;
+            if (item.Type == typeof(TItem).Name)
+                return item;
+
+            return null;
```
IEvangelist, line 52:
```suggestion
            return string.IsNullOrEmpty(item.Type) || item.Type == typeof(TItem).Name
                ? item
                : null;
```

We need to account for the fact that some earlier versions of the SDK didn't have the `.Type` property. So when an `item` is hydrated, it's possible that the `.Type` property is `null`. This is why there is an `.IsDefined()` check on the expression based overload.

IEvangelist (commented):
This looks good, thank you. We'll :shipit: when you make the suggested changes.

IEvangelist:
> Ensures the item returned is of the same type as requested.

Hi @RichMercer - another thought that I had was for this pull request to also include an update to the XML comments on the `GetAsync` overload that has `<remarks>` calling attention to the fact that `null` is returned when the `.Type` is specified but doesn't match.

IEvangelist (approved):

IEvangelist:
@all-contributors please add @RichMercer for code

IEvangelist:
@all-contributors please add @RichMercer for code

IEvangelist:
@all-contributors please add @RichMercer for code

PR #2, opened 2020-10-08
Enhanced to allow for custom partition key paths and variables

gdodd1977:
Added PartitionKeyPath to Options.
Updated IRepository to have overloads that accept custom partitionkey.
Updated Integration tests
Major version since contains breaking changes

### DIFF
--- a/Microsoft.Azure.CosmosRepository.Samples/ServiceTier/Person.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/ServiceTier/Person.cs
@@ -7,7 +7,7 @@ using System;
 
 namespace ServiceTier
 {
-    public class Person : Item
+    public class Person : ItemBase
     {
         public DateTime BirthDate { get; set; }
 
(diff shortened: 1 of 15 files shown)


IEvangelist:
This is a great addition, but I'm actually already working on adding this functionality. I do apologize for not making that clear sooner. I'll be better at adding issues first, and working against those accordingly.

gdodd1977:
No worries. One other thing I was thinking about adding was the ability to add multiple Items to the same container. Not sure if you have something in the works for that too.

IEvangelist:
> No worries. One other thing I was thinking about adding was the ability to add multiple Items to the same container. Not sure if you have something in the works for that too.

That is the default behavior already. When you do not specify to have a `ContainerPerItemType` a single container persists items of any type.

gdodd1977:
So how would you do a mix and match scenario if that setting is on the RepositoryOptions on your ServiceCollectionExtensions that creates everything?

IEvangelist:
> So how would you do a mix and match scenario if that setting is on the RepositoryOptions on your ServiceCollectionExtensions that creates everything?

There is no mixing and matching, it is a mutually exclusive option. You either store all things in a single container - or you place each type in its own container.

PR #8, opened 2020-10-17
Update to .NET Standard 2.1 and clean up

wforney:


wforney:
I also added an EnumerateAsync method using IAsyncEnumerable<TItem> to the repository, among other things. Feel free to pick pieces if you like.

.editorconfig
```diff
@@ -18,16 +18,16 @@ trim_trailing_whitespace = true
...
 
 # this. preferences 
-dotnet_style_qualification_for_field = false:error 
-dotnet_style_qualification_for_property = false:error 
-dotnet_style_qualification_for_method = false:error 
-dotnet_style_qualification_for_event = false:error
+dotnet_style_qualification_for_field = true:error 
+dotnet_style_qualification_for_property = true:error 
+dotnet_style_qualification_for_method = true:error 
+dotnet_style_qualification_for_event = true:error
```
IEvangelist, line 30:
```suggestion
dotnet_style_qualification_for_field = false:error
dotnet_style_qualification_for_property = false:error
dotnet_style_qualification_for_method = false:error
dotnet_style_qualification_for_event = false:error
```

I do not like the `this` qualifier, please leave these as is - it is a style choice that I prefer and a redundancy that I would not like to introduce.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/ExampleService.cs
```diff
@@ -1,37 +1,36 @@
...
-using Microsoft.Azure.CosmosRepository;
+﻿// Copyright (c) IEvangelist. All rights reserved. Licensed under the MIT License.
 
 namespace ServiceTier
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Linq.Expressions;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.CosmosRepository;
```
IEvangelist, line 9: I prefer `using` statements to be outside the namespace, not within - please undo these changes as well.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/ExampleService.cs
```diff
@@ -1,37 +1,36 @@
...
+
     public class ExampleService : IExampleService
     {
-        readonly IRepository<Person> _personRepository;
+        private readonly IRepository<Person> personRepository;
```
IEvangelist, line 13: I prefer to omit default access modifiers, for example `private` on fields - this is again just a style preference.

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosPartitionKeyPathProviderTests.cs
```diff
@@ -28,6 +27,6 @@ public class PickleChipsItem : Item
         [JsonProperty(PropertyName = "pickles")]
         public string PartitionBits { get; set; } = "Hey, where's the chips?!";
 
-        protected override string GetPartitionKeyValue() => PartitionBits;
+        protected override string GetPartitionKeyValue() => this.PartitionBits;
```
IEvangelist, line 30:
```suggestion
        protected override string GetPartitionKeyValue() => PartitionBits;
```

The `this` qualifier is redundant.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerProvider.cs
```diff
@@ -1,23 +1,22 @@
...
 namespace Microsoft.Azure.CosmosRepository.Providers
 {
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Cosmos;
+    using Microsoft.Azure.CosmosRepository.Options;
+
     /// <summary>
-    /// The cosmos container provider exposes a means of providing
-    /// an instance to the configured <see cref="Container"/> object.
+    /// The cosmos container provider exposes a means of providing an instance to the configured
```
IEvangelist, line 10: Did you rely on some formatting tool, or something? Curious as I did not...

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosPartitionKeyPathProvider.cs
```diff
@@ -1,28 +1,28 @@
...
 
+        private static readonly Type partitionKeyNameAttributeType = typeof(PartitionKeyPathAttribute);
+
         /// <inheritdoc />
         public string GetPartitionKeyPath<TItem>() where TItem : Item =>
-            _partionKeyNameMap.GetOrAdd(typeof(TItem), GetPartitionKeyNameFactory);
+            partionKeyNameMap.GetOrAdd(typeof(TItem), GetPartitionKeyNameFactory);
 
-        static string GetPartitionKeyNameFactory(Type type)
+        private static string GetPartitionKeyNameFactory(Type type)
```
IEvangelist, line 22:
```suggestion
        static string GetPartitionKeyNameFactory(Type type)
```

Redundant access modifier.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosPartitionKeyPathProvider.cs
```diff
@@ -1,28 +1,28 @@
...
-        static readonly ConcurrentDictionary<Type, string> _partionKeyNameMap =
+        private static readonly ConcurrentDictionary<Type, string> partionKeyNameMap =
             new ConcurrentDictionary<Type, string>();
 
+        private static readonly Type partitionKeyNameAttributeType = typeof(PartitionKeyPathAttribute);
+
         /// <inheritdoc />
         public string GetPartitionKeyPath<TItem>() where TItem : Item =>
-            _partionKeyNameMap.GetOrAdd(typeof(TItem), GetPartitionKeyNameFactory);
+            partionKeyNameMap.GetOrAdd(typeof(TItem), GetPartitionKeyNameFactory);
```
IEvangelist, line 20:
```suggestion
            _partionKeyNameMap.GetOrAdd(typeof(TItem), GetPartitionKeyNameFactory);
```

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
```diff
@@ -1,96 +1,101 @@
...
+                            Properties.Resources.TheInsertNameHereIsRequiredWhenContainerPerItemTypeIsFalse,
+                            nameof(DefaultCosmosContainerProvider<TItem>.options.DatabaseId)));
                 }
-                if (_options.ContainerId is null)
+
+                if (this.options.ContainerId is null)
                 {
-                    throw new ArgumentNullException(
-                        $"The {nameof(_options.ContainerId)} is required when container per item type is false.");
+                    throw new NullReferenceException(
```
IEvangelist, line 61: I understand the formatting of the string, and the usage of the resource for storing it. Curious about the choice to change the `nameof`, using the `_options.ContainerId` works and is a lot less verbose. Also, you changed the type of the exception here, which is not desirable. I prefer to say that if part of the options are missing, then the argument is `null` as it is part of the parameter that we're validating against.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
```diff
@@ -1,96 +1,101 @@
...
+    /// <inheritdoc />
     internal class DefaultCosmosContainerProvider<TItem>
         : ICosmosContainerProvider<TItem> where TItem : Item
     {
-        readonly Lazy<Task<Container>> _lazyContainer;
-        readonly RepositoryOptions _options;
-        readonly ICosmosClientProvider _cosmosClientProvider;
-        readonly ICosmosPartitionKeyPathProvider _cosmosPartitionKeyPathProvider;
-        readonly ILogger<DefaultCosmosContainerProvider<TItem>> _logger;
+        private readonly ICosmosClientProvider cosmosClientProvider;
```
IEvangelist, line 17: Again, all redundant access modifiers and `this` qualifiers should be reverted. Name private fields with `_` prefix, please.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
```diff
@@ -1,96 +1,101 @@
...
-using Microsoft.Extensions.Options;
+﻿// Copyright (c) IEvangelist. All rights reserved. Licensed under the MIT License.
 
 namespace Microsoft.Azure.CosmosRepository.Providers
 {
-    /// <inheritdoc/>
+    using System;
+    using System.Threading.Tasks;
+
+    using Microsoft.Azure.Cosmos;
```
IEvangelist, line 8: Again, `using` should be at start of file, not within namespace please.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosClientProvider.cs
```diff
@@ -1,43 +1,82 @@
-﻿// Copyright (c) IEvangelist. All rights reserved.
-// Licensed under the MIT License.
-
-using System;
-using System.Threading.Tasks;
-using Microsoft.Azure.Cosmos;
-using Microsoft.Azure.CosmosRepository.Options;
-using Microsoft.Extensions.Options;
+﻿// Copyright (c) IEvangelist. All rights reserved. Licensed under the MIT License.
```
IEvangelist, line 1: This entire file can be reverted to its original version, the dispose pattern isn't needed here since all we're doing is cascading a call to dispose. No reason to maintain state of dispose invocations, provide finalizer, or GC calls, etc. - keep it simple.

Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
```diff
@@ -1,68 +1,57 @@
-﻿// Copyright (c) IEvangelist. All rights reserved.
-// Licensed under the MIT License.
-
-using Microsoft.Azure.Cosmos;
+﻿// Copyright (c) IEvangelist. All rights reserved. Licensed under the MIT License.
```
IEvangelist, line 1: I'm not sure I like this being formatted this way either, the options were structured in an order that made sense - now it is just alphabetical...which is less ideal for authors.

Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Company>Microsoft Corporation</Company>
     <Product>IEvangelist Azure Cosmos Repository</Product>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
IEvangelist, line 6: Unfortunately, I'd prefer to keep this on .NET Standard 2.0 because it is a much broader spectrum of users. .NET Standard 2.1 is limited with who can consume this. While I fully understand and appreciate the effort, and I too am sad about missing out on some of the features - such as async streams and nullable reference types, I'd like to keep this on 2.0 for now.

IEvangelist (changes requested):
Thank you for taking the time to post this pull request, and spend time trying to improve this project. There are many style changes in this PR that are undesirable, and lots of formatting updates that are making it hard to review. Could you please revert some of the following style changes?

-  Alphabetically ordering members
- `this` qualifier
- Removal of `_` prefix on fields
- Explicit access modifier of defaults, for example `private` could be removed

Also, the other big thing is that I would like to stay on version .NET Standard 2.0 for now as it allows this library to be consumed by a much broader set of runtimes.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/Program.cs
```diff
@@ -67,8 +66,10 @@ static async Task RawRepositoryExampleAsync(IRepository<Person> repository)
             _ = await repository.CreateAsync(new[] { maryShaw, calvinWeatherfield });
 
             // Reading...
-            Person mary = await repository.GetAsync(maryShaw.Id, maryShaw.SyntheticPartitionKey);
+            Person? mary = await repository.GetAsync(maryShaw.Id, maryShaw.SyntheticPartitionKey);
+            _ = mary ?? throw new NullReferenceException(nameof(mary));
```
IEvangelist, line 70: Why is this necessary?

Microsoft.Azure.CosmosRepository/src/Attributes/PartitionKeyPathAttribute.cs
```diff
@@ -1,34 +1,31 @@
...
+    /// value will act as the partition key. Partition key paths should start with "/", for example
+    /// "/partition". For more information, see https://docs.microsoft.com/azure/cosmos-db/partitioning-overview.
     /// </summary>
-    /// <remarks>
-    /// By default, "/id" is used.
-    /// </remarks>
+    /// <remarks>By default, "/id" is used.</remarks>
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
-    public class PartitionKeyPathAttribute : Attribute
+    public sealed class PartitionKeyPathAttribute : Attribute
```
IEvangelist, line 16: I like this being `sealed` good suggestion.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -1,162 +1,204 @@
...
-            DeleteAsync(id, new PartitionKey(partitionKeyValue ?? id));
-
-        /// <inheritdoc/>
-        public async ValueTask DeleteAsync(string id, PartitionKey partitionKey)
+        /// <inheritdoc />
+        public async ValueTask<TItem> UpdateAsync(TItem value)
         {
-            ItemRequestOptions options = RequestOptions.Options;
-            Container container = await _containerProvider.GetContainerAsync().ConfigureAwait(false);
+            _ = value ?? throw new ArgumentNullException(nameof(value));
```
IEvangelist, line 179: I'm a bit conflicted about these, but I suppose they do make sense. We should probably add unit tests for these cases.

Microsoft.Azure.CosmosRepository/src/Extensions/ParameterRebinder.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) IEvangelist. All rights reserved. Licensed under the MIT License.
```
IEvangelist, line 1: I like that this was separated into its own file, kudos.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/ExampleService.cs
wforney: There was actually a reason for usings being inside the namespace historically. Here is something that explains: https://stackoverflow.com/questions/125319/should-using-directives-be-inside-or-outside-the-namespace#:~:text=Putting%20it%20inside%20the%20namespaces%20makes%20the%20declarations,whether%20they%20go%20outside%20or%20inside%20the%20namespace.

I have been putting using statements inside namespace for a really long time (10+ years) so I'll have to fight my muscle memory on fixing these if you want it that way.

.editorconfig
wforney: Ok, generally I use the old style guide that was internal to MS a while back and used to be StyleCop oriented. Looks like things have changed a bit since I last updated that. If you use the underscore prefix then you don't need this on everything, but if you don't use the underscore it is making it clear if the variable is local or part of the class.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/ExampleService.cs
wforney: Those were put in by codemaid. I didn't hand enter those so I can flip the switch and revert those easily enough.

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosPartitionKeyPathProviderTests.cs
wforney: The other reason to use this in these situations is to distinguish between a static and an instance variable. It just prevents you from accidentally doing silly things. I'll go flip the config around a bit and reformat things.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerProvider.cs
wforney: Yes, I used the CodeMaid extension and built-in auto-format. It honors editorconfig and whatnot. My main objective was to get it into my standard form, which used to match what Microsoft does internally with their releases. Things have changed there a bit since I put out something there though so I'll have to ping some old friends and see what the newer standards are. I don't think they have guidance on code style across the company any more but I could be wrong.

Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
wforney: Alphabetical is the default setting in CodeMaid. I just ran the formatter on it... It does some opinionated stuff... Most of the changes in this PR are cosmetic. There are some actual things though. I'll go make a version that is formatted closer to the original and incrementally change things. That should be a bit easier to read.

Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
wforney: Ok, fine by me. I'm going to keep my 2.1 version going, with formatting changes to closer match yours. I'll do another PR with just the 2.0 and significant stuff.

Microsoft.Azure.CosmosRepository.Samples/ServiceTier/Program.cs
wforney: GetAsync can return a null so you need the reference nullable Person. The null check after that may just have been me fixing a warning on the next line (not shown), but it would blow up either way without it so I don't suppose it makes a difference.

Microsoft.Azure.CosmosRepository/src/Attributes/PartitionKeyPathAttribute.cs
wforney: That was a code analysis suggestion from VS. :)

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
wforney: Yeah, I put in some of these but I know I didn't catch all of them.

Microsoft.Azure.CosmosRepository/src/Extensions/ParameterRebinder.cs
wforney: 👍

wforney:
I'll close this one and make another one in a bit.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerProvider.cs
IEvangelist: I work for Microsoft, the .NET team does have a common *.editorconfig* but each project can have their own thing. I was following what the Azure Cosmos DB .NET SDK repo was doing as it was my hope that they'd eventually adopt this project.

PR #10, opened 2020-11-06
Support azure function startup

Invvard:
I added an Azure Function oriented AddCosmosRepository() so that I could load it simply from the FunctionStartup.
It comes with an Azure Function sample app that demonstrate how to use the new service injector.


IEvangelist:
This looks awesome so far, I'll take a deeper look later this weekend - thank you @Invvard

Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Model/PostUserRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <param name="postUserRequest">The <see cref="PostUserRequest"/> to start with.</param>
+        public static implicit operator User(PostUserRequest postUserRequest)
+        {
+            return new User()
+                   {
+                       FirstName = postUserRequest.FirstName,
+                       LastName = postUserRequest.LastName,
+                       EmailAddress = postUserRequest.EmailAddress,
+                   };
+        }
```
IEvangelist, line 30:
```suggestion
        public static implicit operator User(PostUserRequest postUserRequest) =>
            new User
            {
                FirstName = postUserRequest?.FirstName,
                LastName = postUserRequest?.LastName,
                EmailAddress = postUserRequest?.EmailAddress,
            };
```

Use an expression bodied member, and remove redundant `()` parenthesis.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosClientProvider.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.CosmosRepository.Providers
     class DefaultCosmosClientProvider : ICosmosClientProvider, IDisposable
     {
         readonly Lazy<CosmosClient> _lazyCosmosClient;
+        readonly ICosmosClientOptionsProvider _cosmosClientOptionsProvider;
```
IEvangelist, line 16: Isn't this is a breaking change? If the consumer the `.AddCosmosRepository(IConfiguration config)` overload (which all users are currently using now), this will break them as this is only added in the Azure function path.

Might be easier to just register your own `CosmosClientOptions` instance externally. I'll have to think about this a bit, and try it out later.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosClientOptionsProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// an instance to the configured <see cref="CosmosClientOptions"/> object,
+    /// which is shared.
+    /// </summary>
+    public interface ICosmosClientOptionsProvider
+    {
+        /// <summary>
+        /// Gets the configured <see cref="CosmosClientOptions"/> instance.
+        /// </summary>
+        /// <returns>A <see cref="CosmosClientOptions"/> instance.</returns>
+        CosmosClientOptions GetClientOptions { get; }
```
IEvangelist, line 16:
```suggestion
        CosmosClientOptions ClientOptions { get; }
```

This shouldn't be named with a "Get" prefix unless it is a function, since it is a property - just name it `ClientOptions`.

Microsoft.Azure.CosmosRepository/test/DefaultRepositoryFactoryTests.cs
```diff
@@ -17,15 +17,17 @@ public class DefaultRepositoryFactoryTests
...
-            IServiceCollection services = new ServiceCollection();
             IConfigurationRoot configuration =
                 new ConfigurationBuilder()
                     .AddInMemoryCollection(new Dictionary<string, string>
-                    {
-                        ["RepositoryOptions:CosmosConnectionString"] = "Testing"
-                    })
+                                           {
+                                               ["RepositoryOptions:CosmosConnectionString"] = "Testing"
+                                           })
```
IEvangelist, line 25:
```suggestion
                    {
                        ["RepositoryOptions:CosmosConnectionString"] = "Testing"
                    })
```

Unnecessary styling change.

IEvangelist (changes requested):
I'm thinking we may want to revisit a few areas, nothing major. What do you think?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosClientProvider.cs
Invvard: I'm not sure I understand. Are you suggesting to create a separate AddCosmosRepository() (named, for example, AddCosmosRepositoryForFunction()) ?
As for externally registering CosmosClientOptions, it is possible to register the RepositoryOptions setting section then extract the IConfiguration instance but it breaks the dependency injection pattern. That's why I did this feature.

In fact, I wanted to make this available for all .NET Core project type but, I'm not sure that's possible due to different ways to store and load settings. Or may be there's a way to ConnectionString, DatabaseId and other settings in the same IConfiguration object but I don't know the ServiceTier and WebTier app type enough.

Invvard:
> This looks awesome so far, I'll take a deeper look later this weekend - thank you @Invvard
Thank you for the kind words and even more for all the work you put into this project of yours !

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosClientProvider.cs
Invvard: Hi @IEvangelist,
Did you have time to think about it ?
If you want, we could talk about it over Teams (now that we all are used to it !), it will be much quickier for you to explain it to me. Anyway, I'm available any time :)

IEvangelist: Sorry for the delay, yes - let's jump on a call quick. Email me, my email is on my GitHub profile @IEvangelist.

Invvard: I just sent you an email :)

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosClientOptionsProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Azure.Cosmos;
+using Microsoft.Azure.CosmosRepository.Options;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.CosmosRepository.Providers
+{
+    /// <summary>
+    /// 
+    /// </summary>
```
IEvangelist, line 13:
```suggestion
    /// <inheritdoc />
```

IEvangelist (approved):

IEvangelist:
@all-contributors please add @Invvard for tests and code

IEvangelist:
@all-contributors please add @Invvard for examples

PR #24, opened 2020-12-02
Revert ConfigurationExtensions version

### LINKED ISSUE #23: Issue loading Microsoft.Extensions.DependencyInjection.Abstractions
I'm not sure if this is a problem with my machine or not, but wanted to ask.

When I try to run my Azure Function after bringing in and using your awesome library, I get a runtime error:

```
[2020-11-27T15:01:03.561Z] Api: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=5.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
Value cannot be null. (Parameter 'provider')
```

I've tried to mirror your sample, so my project file looks almost the same as yours. Here's the most important bits:

```xml
<PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
  </PropertyGroup>
...
<ItemGroup>
    <PackageReference Include="IEvangelist.Azure.CosmosRepository" Version="2.1.1" />
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.10" />
</ItemGroup>
```
The difference is that I fetched your project in NuGet, and in your sample you're using a project reference. When I look at dependencies in NuGet, I see it is supported in .NET Standard 2.0:

image

I'm using .NET Core 3.1, as you are, so that should be OK. Knowing the `Microsoft.Azure.Cosmos` dependency wasn't an issue, I explicitly referenced the `Microsoft.Extensions.*` dependencies in my project file but that didn't help.

I noticed a GitHub issue from a few weeks ago, noting a potential problem [with the host version](https://github.com/Azure/azure-functions-host/issues/6893). Are these `5.0.0` dependencies mucking things up?

daveabrock:
This issue resolves an issue where a version of `Microsoft.Extensions.Options.ConfigurationExtensions` was incompatible with the v3 Azure Functions host. The fix is to change the version from `5.0.0` to `3.1.10`.

Fixes #23

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
+++ b/Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
@@ -45,7 +45,7 @@
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.14.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="3.1.10" />
     <PackageReference Include="MinVer" Version="2.3.1">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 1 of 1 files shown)


Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
```diff
@@ -45,7 +45,7 @@
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.14.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
```
IEvangelist, line 47: Now I'm curious, should these two be bumped down too? Is there a `3.1.10` version for them?

IEvangelist:
@all-contributors please add @daveabrock for code

Issue #26, opened 2020-12-15
[Question] Assigning the PartitionKey


afacanerman:
Hi 

I am trying to use this package configuring the custom partition key (/userIdentifier) instead of /Id, but I am getting an error from cosmos client as below
`System.ArgumentException: The requested partition key path '/id' does not match existing Container 'VideoElapsedTimeData' with partition key path '/userIdentifier' (Parameter 'PartitionKey')`

```
public class VideoElapsedTimeData : Item
    {
        [JsonProperty("userIdentifier")] public string UserIdentifier { get; set; }
        [JsonProperty("consumedTimeInMs")] public double ConsumedTimeInMs { get; set; }
        [JsonProperty("metadata")] public string Metadata { get; set; } = string.Empty;
        [JsonProperty("videoType")] public string VideoType { get; set; } = string.Empty;
        [JsonProperty("viewedPercentage")] public double ViewedPercentage { get; set; }
        [JsonProperty("trimInMs")] public double TrimInMs { get; set; }
        [JsonProperty("trimOutMs")] public double TrimOutMs { get; set; }
        [JsonProperty("timestamp")] public DateTimeOffset Timestamp { get; set; } = DateTimeOffset.UtcNow;

        [JsonProperty("videoId")] public string VideoId { get; set; }

        protected override string GetPartitionKeyValue() => UserIdentifier;
    }
```
My container partition key set as /userIdentifier 

![image](https://user-images.githubusercontent.com/972726/102246695-f2201d80-3ef6-11eb-81af-5f5b1f482f4b.png)

Probably I am missing something I would be so glad if you could help, please.

Thank you

afacanerman:
I think I might have solved the problem with putting

 **[PartitionKeyPath("/userIdentifier")]**
    public class VideoElapsedTimeData : Item
    {


on top of the SubItem and deleting the container and create via insert

IEvangelist:
Yes, the `PartionKeyPath` needs to be specified in conjunction with overriding the `GetPartionKeyValue()`. If that works, we can close this issue. I did add source documentation on this:

- [PartitionKeyPathAttribute](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Attributes/PartitionKeyPathAttribute.cs#L9-L13)
[...]

PR #28, opened 2020-12-16
Fix readme localsettings.json doc

daveabrock:
As presented currently, this does not work (with my Function anyway). When placing it inside of `Values`, as is correctly documented in the Function sample, works like a charm. Fixing document to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,13 +125,13 @@ Depending on the [.NET configuration provider](https://docs.microsoft.com/dotnet
     }
   },
   "AllowedHosts": "*",
-  "RepositoryOptions": {
-    "CosmosConnectionString": "<Your-CosmosDB-ConnectionString>",
-    "DatabaseId": "<Your-CosmosDB-DatabaseName>",
-    "ContainerId": "<Your-CosmosDB-ContainerName>",
-    "OptimizeBandwidth": true,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,13 +125,13 @@ Depending on the [.NET configuration provider](https://docs.microsoft.com/dotnet
...
   },
   "AllowedHosts": "*",
-  "RepositoryOptions": {
-    "CosmosConnectionString": "<Your-CosmosDB-ConnectionString>",
-    "DatabaseId": "<Your-CosmosDB-DatabaseName>",
-    "ContainerId": "<Your-CosmosDB-ContainerName>",
-    "OptimizeBandwidth": true,
-    "ContainerPerItemType": true,
-    "AllowBulkExecution": true
+  "Values": {
```
IEvangelist, line 128: I see what you're saying here, however; this would only be applicable to Azure Functions. You could add another section here, that differentiates between service / web tier and functions - then provide example _appsettings.json_ files. That would make it more explicit.

IEvangelist:
Hey @daveabrock - I'm adding a bit of new README content to provide some example usage to address #26 - I could adjust this during that effort if you'd okay with that?

daveabrock:
Works for me.

PR #36, opened 2021-02-01
IItem abstraction to allow for custom base item classes

### LINKED ISSUE #35: IItem default implementation?
Very cool abstraction, thank you! This is more of a "discussion," but just curious since it's so lightweight, whether you think a C# 8 "IItem" [default interface implementation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods) would suffice as a substitute for the concrete [Item](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Item.cs) class. The use case here would be to support an existing model class hierarchy and hang these "traits" off that implementation. Id and Type currently require a backing field, which I'm not sure how to best accomplish this in that case. Just a thought! :)

dcuccia:
Per [this discussion](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/issues/35), this PR creates an `IItem` abstraction over the Item class, to allow for a custom implementation of the concrete base class when desired. A typical use case is in a brownfield scenario with an existing base entity class, and/or scenarios where base adaptors for multiple databases need to coexist in the same entity. An example usage is shown in _DefaultRepositoryFactoryTests.cs_ [here](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/compare/main...dcuccia:main#diff-ecc142928f54ba144c938a65907992fb12fc8b7d7560fe2d147377728d7fb462). This is primarily to show the usage of the interface - clearly, we might do more to test this, beyond the 1-liner addition to `RepositoryFactoryCorrectlyGetsRepositoryTest()`. Open to suggestions! Thanks for considering!

Closes #35

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Item.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Item.cs
@@ -9,24 +9,6 @@ using System;
 namespace Microsoft.Azure.CosmosRepository
 {
     /// <summary>
-    /// The base interface used for all repository object or object graphs.
-    /// </summary>
-    public interface IItem
-    {
-        /// <summary>
-        /// Gets or sets the item's globally unique identifier.
-        /// </summary>
-        string Id { get; set; }
-
-        /// <summary>
-        /// Gets or sets the item's type name.
-        /// </summary>
-        string Type { get; set; }
-
-        internal PartitionKey PartitionKey { get; }
-    }
-
-    /// <summary>
     /// A base helper class that implements IItem
     /// </summary>
     /// <example>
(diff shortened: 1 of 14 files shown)


Microsoft.Azure.CosmosRepository/src/Item.cs
```diff
@@ -64,4 +68,5 @@ public abstract class Item
         /// <returns>The <see cref="Item.Id"/> unless overridden by the subclass.</returns>
         protected virtual string GetPartitionKeyValue() => Id;
     }
+
```
IEvangelist, line 71:
```suggestion
```

Remove extra new line

IEvangelist (approved):
This is awesome, thank you so much.

IEvangelist:
Closes #35

IEvangelist:
@all-contributors please add @dcuccia for code

dcuccia:
> This is awesome, thank you so much.

Oh great, glad that worked. :)

PR #54, opened 2021-06-18
Returns null instead of 404 exception.

mumby0168:
Have been using the library today and found when doing a point read an the item does not exist rather than getting null back from the repository it throws an exception. Changed it to now return null when a ComosException is thrown with the 404 status code.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
+++ b/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos;
@@ -19,7 +20,7 @@ using Newtonsoft.Json;
 namespace Microsoft.Azure.CosmosRepository
 {
     /// <inheritdoc/>
-    internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : IItem
+    internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : class, IItem
     {
         readonly ICosmosContainerProvider<TItem> _containerProvider;
         readonly IOptionsMonitor<RepositoryOptions> _optionsMonitor;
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @mumby0168 - thank you for this PR. This is a breaking change, as there could have been consumers already using this library who were handling the exception. This changes the behavior and now returns `null`. If the code no longer throws, and `null` is returned consumers who were once handling the exception will now get an `NullReferenceException` with an object not set to an instance of an object error at some later point in the execution. In addition to that concern, this change is only applied to a single call - rather than all of the overloads. I'm not sure I like this change.

- It's a breaking change
- Alters expectations from the API
- Returns `null`
- Doesn't apply to all overloads
- Doesn't update unit tests
- Doesn't update docs

I think for now, it's best for your code to catch the exception and return `null` - do you agree?

mumby0168:
Hi @IEvangelist I understand the point about the breaking change I did consider writing an extension method to get the `GetAsync` call for our use which handles this exception which I can do not a problem.

Just for the future which overloads was you referring to I saw that all the overloads accessed the same `GetAsync` method or did you mean methods such as `DeleteAsync` would throw the same exception? 

Also in terms of tests I did not see any tied to this class have I missed them somewhere? 

Final one, where does the documentation live for future reference.

Thanks again,
Billy

IEvangelist:
> Just for the future which overloads was you referring to I saw that all the overloads accessed the same `GetAsync` method or did you mean methods such as `DeleteAsync` would throw the same exception?

No, not the `DeleteAsync` but all the `GetAsync` overloads - there are three total - but only one was updated. Also adding the `class` generic constraint would need to be done on the interface as well. Probably best for parity (and API consistency) sake to do the same to the `GetByQueryAsync` methods.

> Also in terms of tests I did not see any tied to this class have I missed them somewhere?

You're probably correct. I do need to work on writing more tests (in my head they existed already) 😬.

> Final one, where does the documentation live for future reference?

The README.md for now, if time permits I'd love to add more - perhaps a Wiki, or even a standalone site - but time is something I find myself never having enough of.

Should we close this?

mumby0168:
Yes not to worry thank you for your quick responses much appreciated 👍

IEvangelist:
> Yes not to worry thank you for your quick responses much appreciated 👍

I'd love for more contributions to this project. If you have ideas or changes you'd like to propose in the future, please feel free to start with an issue - so we can discuss it. Thanks again.

philip-reed:
Could maybe make this behavior something that consumers opt in to.  Have a configuration value which defaults to the current behavior so exceptions continue to be thrown.  Something like `options.ReturnNullFor404 = true`?

IEvangelist:
@all-contributors please add @mumby0168 for feature and code

PR #57, opened 2021-06-26
add new method to support pagination

### LINKED ISSUE #53: Feature request: pagination support
When we get items from repo, if the count of items is more than several thousands, a better way is to get data by page, can we support it?

zhangzunke:
Sir, I added a method to support pagination, please review it. thanks.

Fixes #53


IEvangelist:
This is awesome, thank you so much @zhangzunke - I'll do a full review this weekend. Thanks again

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.Cosmos.Linq;
+using Microsoft.Azure.CosmosRepository.Dtos;
```
IEvangelist, line 12:
```suggestion
using Microsoft.Azure.CosmosRepository.Pagination;
```

I prefer the namespace of Pagination, instead of Dtos. Let's change this for scoping all of the pagination related features - as Dtos stands for "data transfer object" and doesn't apply directly to pagination.

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResult.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Azure.CosmosRepository.Dtos
+{
+    /// <inheritdoc />
+    public class PagedResult<T> : ListResult<T>
+    {
+        /// <inheritdoc />
+        public long TotalCount { get; set; }
```
IEvangelist, line 9:
```suggestion
        public long TotalCount { get; internal set; }
```

Do not let consumers set this value, we'll manage it internally.

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResult.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 3:
```suggestion
namespace Microsoft.Azure.CosmosRepository.Pagination
```

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResult.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
```
IEvangelist, line 1: Missing copy right header, through out all new files.

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResult.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <inheritdoc />
+        public long TotalCount { get; set; }
+
+        /// <summary>
+        /// Creates a new <see cref="PagedResult{T}"/> object.
+        /// </summary>
+        public PagedResult()
+        {
+
+        }
```
IEvangelist, line 17:
```suggestion
        /// <summary>
        /// Creates a new <see cref="PagedResult{T}"/> object.
        /// </summary>
        public PagedResult() { }
```

Nit: for empty `.ctor` inline declaration to avoid multiple lines.

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResult.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Creates a new <see cref="PagedResult{T}"/> object.
+        /// </summary>
+        /// <param name="totalCount">Total count of Items</param>
+        /// <param name="items">List of items in current page</param>
+        public PagedResult(long totalCount, IReadOnlyList<T> items)
+            : base(items)
+        {
+            TotalCount = totalCount;
+        }
```
IEvangelist, line 28:
```suggestion
            : base(items) =>
            TotalCount = totalCount;
```

Prefer expression body constructors where possible.

Microsoft.Azure.CosmosRepository/src/Dtos/PagedResultRequest.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 1:
```suggestion
// Copyright (c) IEvangelist. All rights reserved.
// Licensed under the MIT License.
﻿
﻿﻿﻿namespace Microsoft.Azure.CosmosRepository.Pagination
```

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos;
+using Microsoft.Azure.CosmosRepository.Dtos;
```
IEvangelist, line 10:
```suggestion
using Microsoft.Azure.CosmosRepository.Pagination;

```

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -61,6 +62,17 @@ public interface IRepository<TItem> where TItem : IItem
...
+        /// <summary>
+        /// Gets an <see cref="PagedResult{TItem}"/> of <see cref="IItem"/>
+        /// </summary>
+        /// <param name="predicate">The expression used for evaluating a matching item.</param>
+        /// <param name="pagedResultRequest">The paging object.</param>
+        /// <param name="cancellationToken">The cancellation token to use when making asynchronous operations.</param>
+        /// <returns>A paged result of item instances who meet the <paramref name="predicate"/> and <paramref name="pagedResultRequest"/> condition.</returns>
+        Task<PagedResult<TItem>> GetListAsync<TPagedRequest>(
+            Expression<Func<TItem, bool>> predicate, TPagedRequest pagedResultRequest,
+            CancellationToken cancellationToken = default);
```
IEvangelist, line 74:
```suggestion
        Task<PagedResult<TItem>> GetListAsync<TPagedRequest>(
            Expression<Func<TItem, bool>> predicate,
            TPagedRequest pagedResultRequest,
            CancellationToken cancellationToken = default);
```

Apply consistent wrapping of parameters.

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 3:
```suggestion
namespace Microsoft.Azure.CosmosRepository.Pagination
```

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public class ListResult<T>
+    {
+        /// <summary>
+        /// Creates a new <see cref="ListResult{T}"/> object.
+        /// </summary>
+        public ListResult()
+        {
+
+        }
```
IEvangelist, line 16:
```suggestion
        /// <summary>
        /// Creates a new <see cref="ListResult{T}"/> object.
        /// </summary>
        public ListResult() { }
```

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <inheritdoc />
+        public IReadOnlyList<T> Items
+        {
+            get { return _items ??= new List<T>(); }
+            set { _items = value; }
+        }
+
+        private IReadOnlyList<T> _items;
```
IEvangelist, line 34:
```suggestion
        public IReadOnlyList<T> Items { get; internal set; } = new List<T>();
```

Prefer auto-properties with initializer, also use `internal set`.

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ListResult{T}"/> object.
+        /// </summary>
+        /// <param name="items">List of items</param>
+        public ListResult(IReadOnlyList<T> items)
+        {
+            Items = items;
+        }
```
IEvangelist, line 25:
```suggestion
        public ListResult(IReadOnlyList<T> items) =>Items = items;
```

Microsoft.Azure.CosmosRepository/src/Dtos/LimitedResultRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 1:
```suggestion
﻿namespace Microsoft.Azure.CosmosRepository.Pagination
```

Also, need to add the copyright header.

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
```
IEvangelist, line 1: Need to add the copyright header.

Microsoft.Azure.CosmosRepository/src/Dtos/IPagedResultRequest.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// This interface is defined to standardize to request a paged result.
+    /// </summary>
+    public interface IPagedResultRequest : ILimitedResultRequest
+    {
+        /// <summary>
+        /// Skip count (beginning of the page).
+        /// </summary>
+        int SkipCount { get; set; }
```
IEvangelist, line 11:
```suggestion
        int SkipCount { get; }
```

Hide set to consumers, the implementation can still assign to it - but we do not want consumers setting this.

Microsoft.Azure.CosmosRepository/src/Dtos/IPagedResultRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 1:
```suggestion
﻿namespace Microsoft.Azure.CosmosRepository.Pagination
```

Need the copyright header.

Microsoft.Azure.CosmosRepository/src/Dtos/ILimitedResultRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.Azure.CosmosRepository.Dtos
```
IEvangelist, line 1:
```suggestion
﻿namespace Microsoft.Azure.CosmosRepository.Pagination
```

Need the copyright header.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -98,6 +99,56 @@ await container.ReadItemAsync<TItem>(id, partitionKey, cancellationToken: cancel
...
+            {
+                foreach (TItem result in await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false))
+                {
+                    results.Add(result);
+                }
+            }
+            return new PagedResult<TItem>(totalCount, results);
+        }
+
+        protected virtual IQueryable<TItem> ApplyPaging<TPagedRequest>(IQueryable<TItem> query, TPagedRequest pagedResultRequest)
```
IEvangelist, line 134: Why is this `protected virtual`? What is the plan for this?

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -98,6 +99,56 @@ await container.ReadItemAsync<TItem>(id, partitionKey, cancellationToken: cancel
             return results;
         }
 
+        /// <inheritdoc/>
+        public async Task<PagedResult<TItem>> GetListAsync<TPagedRequest>(
+            Expression<Func<TItem, bool>> predicate, TPagedRequest pagedResultRequest,
+            CancellationToken cancellationToken = default)
```
IEvangelist, line 105:
```suggestion
        public async Task<PagedResult<TItem>> GetListAsync<TPagedRequest>(
            Expression<Func<TItem, bool>> predicate,
            TPagedRequest pagedResultRequest,
            CancellationToken cancellationToken = default)
```

IEvangelist (changes requested):
Here is a few initial observations. I do really like what you have thus far -- would you be able to add some tests, or a sample app usage? See the samples and test directories.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
zhangzunke: Maybe someone can override the method to add customize pagination.

Microsoft.Azure.CosmosRepository/src/Dtos/IPagedResultRequest.cs
zhangzunke: Actually, the consumers will set the property when they want to pagination, below is the demo code,  maybe I am wrong.
```
var pagedResultRequest = new PagedResultRequest { SkipCount = input.SkipCount, MaxResultCount = input.MaxResultCount };
var mediaList = await _mediaRepository.GetListAsync(predicate, pagedResultRequest);
```

Microsoft.Azure.CosmosRepository/src/Dtos/ListResult.cs
zhangzunke: Good catch

zhangzunke:
Hi @IEvangelist, 
Thanks for your suggestion, and the changes ready for review.

IEvangelist (commented):
Also, from the design perspective - as a consumer of this API how would I know if there are more pages, and is there a way to also communicate the total number of pages? If I query for a max of 10 items, and get back 10 items, how would I know if there are more items than the 10 I received for the first page?

Microsoft.Azure.CosmosRepository/src/Pagination/LimitedResultRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.Azure.CosmosRepository.Pagination
+{
+    /// <inheritdoc />
+    public class LimitedResultRequest : ILimitedResultRequest
+    {
+        /// <summary>
+        /// Default value: 10.
+        /// </summary>
+        public static int DefaultMaxResultCount { get; set; } = 10;
```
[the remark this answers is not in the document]
zhangzunke: Because it is a global setting, if a consumer wants to return a customize count, they can modify the DefaultMaxResultCount  value when starting application,

zhangzunke:
> Hi again @zhangzunke - thanks for all these updates, I have a few more things I'm curious about.
> 
> Also, from the design perspective - as a consumer of this API how would I know if there are more pages, and is there a way to also communicate the total number of pages? If I query for a max of 10 items, and get back 10 items, how would I know if there are more items than the 10 I received for the first page?

Good catch, the method GetListAsync will return a PagedResult<T>, it has a property TotalCount which is all of the items count when you query, so a consumer can use the property to calculate how many pages, the property is not the current page items count.

Microsoft.Azure.CosmosRepository/src/Pagination/LimitedResultRequest.cs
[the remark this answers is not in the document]
IEvangelist: Is `10` a reasonable default? Why not `50`, or `100`?

IEvangelist:
@zhangzunke  Another thought is that the `PagedResult` should also include some calculated properties about the current page and the total number of pages.

zhangzunke:
Hi @IEvangelist, the first design is that the method only provides a basic pagination feature, the pagination method only accept `SkipCount `and `MaxResultCount` parameters,  so you mention the current page and a total number of pages value should be put on consumer side to implementation, that why the PagedResult only gives a total count property to the consumer.

Regarding how to get the next page, because the consumer knows the grid current page, page size and the total count of the data (it comes from `PagedResult` property), so it is very easy to get the next page.

IEvangelist:
> Hi @IEvangelist, the first design is that the method only provides a basic pagination feature, the pagination method only accept `SkipCount `and `MaxResultCount` parameters, so you mention the current page and a total number of pages value should be put on consumer side to implementation, that why the PagedResult only gives a total count property to the consumer.
> 
> Regarding how to get the next page, because the consumer knows the grid current page, page size and the total count of the data (it comes from `PagedResult` property), so it is very easy to get the next page.


@zhangzunke Could you show me an example of it being used, so that I can wrap my head around the expected usage? Do you have an example, where you there are 1,037 records that match the query -- and you page them with a max of 50 per page... what would that look like?

zhangzunke:
Hi @IEvangelist, I showed a simple example as below, in our project these codes should be implemented by typescript on the SPA, so I just showed the basic cases when we used pagination, is that expected usage for you?
```
  private int totalPageCount = 0;
  private int pageSize = 50;
  private int currentPage = 1;
  private IReadOnlyList<Media> mediasDataSource;
  public async Task Demo()
  {
      // Click next button
      await ClickNextButtonAsync();

      // click previous button
      await ClickPreviousButtonAsync();
  }

  private async Task ClickNextButtonAsync()
  {
      currentPage++;
      if (currentPage <= totalPageCount)
      {
          await ClickPreviousOrNextButtonAsync(currentPage);
      }
  }

  private async Task ClickPreviousButtonAsync()
  {
      currentPage--;
      if (currentPage > 0)
      {
          await ClickPreviousOrNextButtonAsync(currentPage + 1);
      }
  }

  private async Task<PagedResult<Media>> PaginationDemo(int skipCount = 0, int maxResultCount = 50)
  {
      var pagedResultRequest = new PagedResultRequest { SkipCount = skipCount, MaxResultCount = maxResultCount };
      Expression<Func<Media, bool>> predicate = x => x.UserId == "xxxxxxx";
      var mediaList = await _mediaRepository.GetListAsync(predicate, pagedResultRequest);
      return mediaList;
  }

  private async Task ClickPreviousOrNextButtonAsync(int currentPage)
  {
      var mediaList = await PaginationDemo((currentPage - 1) * pageSize, pageSize);
      mediasDataSource = mediaList.Items;
      // get total page count, the value will be 21
      totalPageCount = (int)Math.Ceiling(Convert.ToDouble((decimal)mediaList.TotalCount / pageSize));
  }
```

Microsoft.Azure.CosmosRepository/src/Pagination/IPagedResultRequest.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.CosmosRepository.Pagination
+{
+    /// <summary>
+    /// This interface is defined to standardize to request a paged result.
+    /// </summary>
+    public interface IPagedResultRequest : ILimitedResultRequest
```
roji, line 9: The interface inheritance seems to mean that you don't allow skipping (OFFSET) without also taking (LIMIT), although Cosmos does seem to support that. For example, it would be possible to express "give me all elements, skipping the first 100".

zhangzunke: Thank you for the review, the current design does not support the scenario, we can support it easily in the future, we just add another interface `ISkipResultRequest`, and does not inherit the `ILimitedResultRequest`, then add a new branch in the below method. In my opinion, most of the scenarios need the max result count, any suggestions?
```
protected virtual IQueryable<TItem> ApplyPaging<TPagedRequest>(IQueryable<TItem> query, TPagedRequest pagedResultRequest)
{
        //Try to use paging if available
        if (pagedResultRequest is IPagedResultRequest pagedInput)
        {
            return query.Skip(pagedInput.SkipCount).Take(pagedInput.MaxResultCount);
        }

        //Try to limit query result if available
        if (pagedResultRequest is ILimitedResultRequest limitedInput)
        {
            return query.Take(limitedInput.MaxResultCount);
        }

        //Try to limit query result if available
        if (pagedResultRequest is ISkipedResultRequest skipInput)
        {
            return query.Skip(skipInput));
        }

        //No paging
        return query;
}
```

IEvangelist: I think @roji is on to something here, perhaps we should consider this a bit further. I want to make sure this is right. Also, I was informed that the Cosmos DB team actually suggestions continuation tokens instead of the Skip / Take approach to pagination.

zhangzunke: Hi @IEvangelist, I added some code to support the scenario accordingly, now we can skip the specific count and get all the items.
About the continuation tokens, I noticed the document from the Cosmos DB team, the solution is better than the skip/take the approach, I need to more effort on it since I don't familiar with the knowledge, can we merge the PR first, then we can add enhancement to support the continuation tokens in the future.

IEvangelist: That's the thing, I'm not sure the API will be the same. I'm hesitant to introduce a new API that may not be correct and might change with the next PR. I'd rather do it right the first time. I think what we have is awesome, but I want to explore the continuation approach first.

zhangzunke: @IEvangelist  I totally agree with what you said doing it right the first time. I will close the PR and continue to explore the continuation approach in the future. currently, I will use the branch DLL in my project first, I can replace the branch DLL with official NuGet if the continuation approach is available in the future, thank you to discuss with me about the feature.

IEvangelist: I really appreciate it, thank you. Sorry to push back, but I just want to make sure it's the right solution since more people are starting to use this library. In fact, would you be interested in helping maintain it with me? Now that I'm starting to write a book, I have less time to give this project the full attention it deserves. I'd love 💜 more help.

zhangzunke: @IEvangelist I am glad you can invite me to maintain the project with you, but I am afraid I can not take the responsibility since I am a newbie in Cosmos DB, however, I will pay attention to the library and contribute my efforts if I can, thank you again.

zhangzunke:
Cosmos DB team suggest using continuation tokens instead of the Skip/Take approach to pagination, so I closed the PR.

> The RU charge of a query with OFFSET LIMIT will increase as the number of terms being offset increases. For queries that have multiple pages of results, we typically recommend using continuation tokens. Continuation tokens are a "bookmark" for the place where the query can later resume. If you use OFFSET LIMIT, there is no "bookmark". If you wanted to return the query's next page, you would have to start from the beginning.

roji:
Quoting @AndriySvyryd here, the EF Cosmos provider automatically uses continuation tokens under the hood when normal query results are enumerated. In other words, for optimal pagination, simply execute your query and [stream the results](https://docs.microsoft.com/en-us/ef/core/performance/efficient-querying#buffering-and-streaming) (e.g. with foreach); the provider will fetch results incrementally. Avoid buffering the entire results e.g. with ToList. https://github.com/dotnet/efcore/issues/24513 tracks allowing the user to set the number of items fetched each time.

Using Skip may still be useful if one doesn't want to traverse the first N items; but executing the same query with different Skip/Take windows would result in bad perf and increased costs.

(@AndriySvyryd will correct if I wrote anything wrong)

IEvangelist:
@all-contributors please add @zhangzunke for ideas and code

IEvangelist:
@all-contributors please add @roji for research

IEvangelist:
@all-contributors please add @roji for research

IEvangelist:
@all-contributors please add @roji for review

mumby0168:
> Quoting @AndriySvyryd here, the EF Cosmos provider automatically uses continuation tokens under the hood when normal query results are enumerated. In other words, for optimal pagination, simply execute your query and [stream the results](https://docs.microsoft.com/en-us/ef/core/performance/efficient-querying#buffering-and-streaming) (e.g. with foreach); the provider will fetch results incrementally. Avoid buffering the entire results e.g. with ToList. [dotnet/efcore#24513](https://github.com/dotnet/efcore/issues/24513) tracks allowing the user to set the number of items fetched each time.
> 
> Using Skip may still be useful if one doesn't want to traverse the first N items; but executing the same query with different Skip/Take windows would result in bad perf and increased costs.
> 
> (@AndriySvyryd will correct if I wrote anything wrong)

Hi I have been trying to wrap my head around the refresh tokens would this comment on the open issue sound about right, continuation tokens are great for a *load more ...* scenario but for someone who wants to skip directly to page 6 for example they would not have any real benefit?

https://github.com/IEvangelist/azure-cosmos-dotnet-repository/issues/53#issuecomment-892921728

Thanks in advance :)

roji:
@mumby0168 if my mental model is correct (not an expert), you can still use Skip to skip directly to page 6, and from that point, stream everything else (using continuation tokens). IIRC doing so wouldn't reduce Cosmos processing costs for the 5 skipped pages, but would save you from having to transfer that data over the network.

mumby0168:
> @mumby0168 if my mental model is correct (not an expert), you can still use Skip to skip directly to page 6, and from that point, stream everything else (using continuation tokens). IIRC doing so wouldn't reduce Cosmos processing costs for the 5 skipped pages, but would save you from having to transfer that data over the network.

Got it, put the basis of my plan together for the implementation if you could spare some time to have a look it would be greatly appreciated 👍 

https://github.com/IEvangelist/azure-cosmos-dotnet-repository/pull/82

PR #62, opened 2021-07-17
Fix #52

### LINKED ISSUE #52: Data loss with nuget package versions 2.1.4 through to 2.1.9
Sorry for being melodramatic, but this bug is critical enough and will definitely cause data loss in production.  It is likely people will review this when trying to resolve.

Temporary resolution: Downgrade to nuget package version 2.1.3 until this issue has been resolved.  I have verified version 2.1.3 has the correct behaviour.

With versions 2.1.4 through to 2.1.9, if you are storing multiple types within a single container in cosmos (which would be the default for most users), you will quickly corrupt your database.  In summary, the codebase attempts to add an additional where clause to the predicate you supply to GetAsync, but the logic is not correct.  

If you want a work around this bug and use more recent nuget package versions, you MUST include your own type filter within the predicate, for example if your type is MyName, you would have written this:
```
db.GetAsync(i => i.FirstName == "Fred")
```
If you use nuget package 2.1.4 to 2.1.9 you MUST include your own type filter in your supplied predicate like this:
```
db.GetAsync(i => i.FirstName == "Fred" && i.Type == nameof(MyType))
```
If you do not do this on later package version, you will end up corrupting your data.

jrwmacedo:
The property "type" into the IItem interface has been annotated using the JSONProperty attribute and this way, the issue found in the compose method was solved.
A new example using multiple types into the same container was created to test it, it is a bit hard unit test this kind of issue.

### DIFF
--- a/Microsoft.Azure.CosmosRepository.sln
+++ b/Microsoft.Azure.CosmosRepository.sln
@@ -30,6 +30,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AzureFunctionTier", "Micros
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "Samples", "{2EEE3D0D-6457-46D8-B8E8-FA7364942F1F}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebTierMultipleTypesInContainer", "WebTierMultipleTypesInContainer\WebTierMultipleTypesInContainer.csproj", "{A6B41095-3374-4A60-94BF-54C3ACCA5525}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -56,6 +58,10 @@ Global
 		{FD8C66BD-D45F-4D99-891B-0C490E2A8072}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{FD8C66BD-D45F-4D99-891B-0C490E2A8072}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{FD8C66BD-D45F-4D99-891B-0C490E2A8072}.Release|Any CPU.Build.0 = Release|Any CPU
+		{A6B41095-3374-4A60-94BF-54C3ACCA5525}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{A6B41095-3374-4A60-94BF-54C3ACCA5525}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{A6B41095-3374-4A60-94BF-54C3ACCA5525}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{A6B41095-3374-4A60-94BF-54C3ACCA5525}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -64,6 +70,7 @@ Global
 		{AE98C44F-4F35-4AF8-9141-1DBCB00468C8} = {2EEE3D0D-6457-46D8-B8E8-FA7364942F1F}
 		{11385AFD-CEA1-47EB-847A-D29D2F929F74} = {2EEE3D0D-6457-46D8-B8E8-FA7364942F1F}
 		{FD8C66BD-D45F-4D99-891B-0C490E2A8072} = {2EEE3D0D-6457-46D8-B8E8-FA7364942F1F}
+		{A6B41095-3374-4A60-94BF-54C3ACCA5525} = {2EEE3D0D-6457-46D8-B8E8-FA7364942F1F}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {6AAE7641-B62C-48BA-8FE6-0F819E5B45EF}
--- a/Microsoft.Azure.CosmosRepository/src/IItem.cs
+++ b/Microsoft.Azure.CosmosRepository/src/IItem.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) IEvangelist. All rights reserved.
 // Licensed under the MIT License.
 
+using Newtonsoft.Json;
+
 namespace Microsoft.Azure.CosmosRepository
 {
     /// <summary>
@@ -16,6 +18,7 @@ namespace Microsoft.Azure.CosmosRepository
         /// <summary>
         /// Gets or sets the item's type name.
         /// </summary>
+		[JsonProperty("type")]
         string Type { get; set; }
 
         /// <summary>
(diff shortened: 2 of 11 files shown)


IEvangelist:
Hi @jrwmacedo - thanks for this PR. I think we should change this a bit and expose the Cosmos DB serialization options to consumers so they can control this. With this change, we could make the default options be case insensitive - thus solving our problem here too. But if we were to decorate this, it's set that way forever.

<https://docs.microsoft.com/dotnet/api/microsoft.azure.cosmos.cosmosclientoptions.serializer>

What do you think about that?

IEvangelist:
I exposed lots of settings now to consumers, see: 6b0e1b0803945150b6e585f823fb8e7bc95a932a and 64271f6c44af6be6170073b663c09e209b38db2d. We might be able to close this one now, thoughts?

jrwmacedo:
> I exposed lots of settings now to consumers, see: [6b0e1b0](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/commit/6b0e1b0803945150b6e585f823fb8e7bc95a932a) and [64271f6](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/commit/64271f6c44af6be6170073b663c09e209b38db2d). We might be able to close this one now, thoughts?

Hey @IEvangelist , Thanks for that man, I agree with your suggestion make more sense. I would change that later but as you already did, I appreciate it.

IEvangelist:
Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?

jrwmacedo:
> Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?

I will do that...Give me a couple of hours and I will let you know if it is working properly for me.

IEvangelist:
> > Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?
> 
> I will do that...Give me a couple of hours and I will let you know if it is working properly for me.

Thank you so much! One test scenario would be to create some objects with the https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Options/RepositorySerializationOptions.cs#L31 set to `Default`, this would save them with non-camel case. Then change your settings, to use `CamelCase` and see if there is a loss of data.

jrwmacedo:
> > > Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?
> > 
> > 
> > I will do that...Give me a couple of hours and I will let you know if it is working properly for me.
> 
> Thank you so much! One test scenario would be to create some objects with the https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Options/RepositorySerializationOptions.cs#L31 set to `Default`, this would save them with non-camel case. Then change your settings, to use `CamelCase` and see if there is a loss of data.

Hi again @IEvangelist ,
After doing some tests, I faced an odd situation here...
The CosmosClient is always assuming the serialization as CamelCase and it is ignoring our configuration setting.
I tried to figure out what is going on but I couldn't at the moment. I thought the configuration binder was not working well or something related to the lazy loading but it is weird because the SerializerConfiguration is applied accordingly based on the debug values.
The changes solved the type issue because now is always assuming the CamelCase as default but in another hand, we have the setting ignored if someone need to change that manually as you proposed.
I tried to record this gif to expose and clarify the situation.
https://drive.google.com/file/d/1uAGCdC94XSaeu9nmZddcvJQE6v9tqUBX/view?usp=sharing

Any thoughts ?

Cheers!

IEvangelist:
> > > > Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?
> > > 
> > > 
> > > I will do that...Give me a couple of hours and I will let you know if it is working properly for me.
> > 
> > 
> > Thank you so much! One test scenario would be to create some objects with the https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Options/RepositorySerializationOptions.cs#L31 set to `Default`, this would save them with non-camel case. Then change your settings, to use `CamelCase` and see if there is a loss of data.
> 
> Hi again @IEvangelist ,
> After doing some tests, I faced an odd situation here...
> The CosmosClient is always assuming the serialization as CamelCase and it is ignoring our configuration setting.
> I tried to figure out what is going on but I couldn't at the moment. I thought the configuration binder was not working well or something related to the lazy loading but it is weird because the SerializerConfiguration is applied accordingly based on the debug values.
> The changes solved the type issue because now is always assuming the CamelCase as default but in another hand, we have the setting ignored if someone need to change that manually as you proposed.
> I tried to record this gif to expose and clarify the situation.
> https://drive.google.com/file/d/1uAGCdC94XSaeu9nmZddcvJQE6v9tqUBX/view?usp=sharing
> 
> Any thoughts ?
> 
> Cheers!

Yes, now I'm curious - did you remove the `[JsonProperty("type")]` attribute decorator on the property before you compiled and tested? That might be the reason the `"type"` name is used?

jrwmacedo:
> > > > > Should we close this PR, or do you want to update it - pull the latest bits and see if it works without your `JsonProperty` attr?
> > > > 
> > > > 
> > > > I will do that...Give me a couple of hours and I will let you know if it is working properly for me.
> > > 
> > > 
> > > Thank you so much! One test scenario would be to create some objects with the https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/Options/RepositorySerializationOptions.cs#L31 set to `Default`, this would save them with non-camel case. Then change your settings, to use `CamelCase` and see if there is a loss of data.
> > 
> > 
> > Hi again @IEvangelist ,
> > After doing some tests, I faced an odd situation here...
> > The CosmosClient is always assuming the serialization as CamelCase and it is ignoring our configuration setting.
> > I tried to figure out what is going on but I couldn't at the moment. I thought the configuration binder was not working well or something related to the lazy loading but it is weird because the SerializerConfiguration is applied accordingly based on the debug values.
> > The changes solved the type issue because now is always assuming the CamelCase as default but in another hand, we have the setting ignored if someone need to change that manually as you proposed.
> > I tried to record this gif to expose and clarify the situation.
> > https://drive.google.com/file/d/1uAGCdC94XSaeu9nmZddcvJQE6v9tqUBX/view?usp=sharing
> > Any thoughts ?
> > Cheers!
> 
> Yes, now I'm curious - did you remove the `[JsonProperty("type")]` attribute decorator on the property before you compiled and tested? That might be the reason the `"type"` name is used?

Morning @IEvangelist , my bad here while I was testing another scenario I reset my branch and I forgot to remove the JSON property (again). It was 2:00 AM (Lisbon time) when I was testing that man! LoL. 
The scenario that I was testing is:
Using the configuration file (appSettings.json) is perfectly fine but if we set the configuration directly in the code like below the option is ignored because the config takes precedence.
`     services.AddCosmosRepository(option =>
            {
                option.DatabaseId = "samples";
                option.ContainerId = "container";
                option.SerializationOptions = new RepositorySerializationOptions
                {
                    PropertyNamingPolicy = CosmosPropertyNamingPolicy.Default
                };
            });
`
Anyway, using the appSettings to handle that is enough to change the default behaviour if necessary, the issue below can be handle later in my opinion.

I am happy to close this PR if you are as well.

IEvangelist:
@all-contributors please add @jrwmacedo for code and ideas

PR #77, opened 2021-07-30
Alternate Configuration for Items Container & Partition Key

### LINKED ISSUE #71: Alternate method to configure container-name/partition-key for an item type
`n/a`

Another way to configure the container name & partition key for a given item type this could be done when configuring the cosmos repository options or another extension method to services. Maybe something like below:

```csharp
services.AddCosmosRepository().AddItem<Person>("people", "/email");
```

This may be better to be done as part of a configure action for this library such as:
```csharp
services.AddCosmosRepository(configure => 
{
    configure.ConfigureRepository<Person>("people", "/email");
}
```

The alternate is already in place currently using the attributes.

**Additional context**
This could reduce the need for reflection using the attribute based approach if this option is used.

mumby0168:
Closes #71

This is still very much open for discussion would love some opinions on this I really like alternative of using attributes on the objects.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.CosmosRepository.Builders
     public interface IItemContainerBuilder
     {
         /// <summary>
-        ///
+        /// The list of already configured options for a <see cref="IItem"/>
         /// </summary>
         IReadOnlyList<ContainerOptions> Options { get; }
 
--- a/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
@@ -1,0 +1,28 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Microsoft.Azure.CosmosRepository.Options;
+
+namespace Microsoft.Azure.CosmosRepository.Builders
+{
+    /// <summary>
+    /// A builder to configure a container for an item
+    /// </summary>
+    public interface IItemContainerBuilder
+    {
+        /// <summary>
+        ///
+        /// </summary>
+        ConcurrentDictionary<Type, ContainerOptions> Options { get; }
+
+        /// <summary>
+        /// Provides a <see cref="ContainerOptions"/> instance to configure a container for an item
+        /// </summary>
+        /// <typeparam name="TItem">The type of <see cref="IItem"/> to configure.</typeparam>
+        /// <returns>Instance of <see cref="IItemContainerBuilder"/></returns>
+        IItemContainerBuilder Configure<TItem>(Action<ContainerOptions> configure) where TItem : IItem;
+    }
+}
--- a/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
@@ -16,13 +16,13 @@ namespace Microsoft.Azure.CosmosRepository.Builders
         /// <summary>
         ///
         /// </summary>
-        ConcurrentDictionary<Type, ContainerOptions> Options { get; }
+        IReadOnlyList<ContainerOptions> Options { get; }
 
         /// <summary>
         /// Provides a <see cref="ContainerOptions"/> instance to configure a container for an item
         /// </summary>
         /// <typeparam name="TItem">The type of <see cref="IItem"/> to configure.</typeparam>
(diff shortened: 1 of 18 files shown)


bradwestness (approved):

Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Azure.CosmosRepository.Builders
+{
+    /// <summary>
+    /// A builder to configure a container for an item
+    /// </summary>
+    public interface IItemContainerBuilder
+    {
+        /// <summary>
+        ///
```
IEvangelist, line 17: We should probably have something here, especially since it is `public`.

Microsoft.Azure.CosmosRepository/src/Options/ContainerOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <summary>
+    /// Options for a container
+    /// </summary>
+    public class ContainerOptions
+    {
+        /// <summary>
+        /// The <see cref="IItem"/> type the container options are for
+        /// </summary>
+        internal Type Type { get; }
+
```
IEvangelist, line 17:
```suggestion
```

Remove extra line

Microsoft.Azure.CosmosRepository/src/Options/ContainerOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        /// <param name="name">The name of the container</param>
+        /// <returns>Instance of <see cref="ContainerOptions"/></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public ContainerOptions WithContainer(string name)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+            return this;
+        }
+
```
IEvangelist, line 49:
```suggestion
```

Remove extra line

Microsoft.Azure.CosmosRepository/src/Options/ContainerOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="ContainerOptions"/>
+        /// </summary>
+        /// <param name="type">The type of <see cref="IItem"/> the options are for</param>
+        public ContainerOptions(Type type)
+        {
+            Type = type;
+        }
```
IEvangelist, line 26:
```suggestion
        public ContainerOptions(Type type) => Type = type;
```

IEvangelist (commented):
I really like this! Looking great, but I'm curious what happens if you do not call the `.WithContainer` and `.WithPartitionKey` -- what is the behavior then? Also, what if you only call one or the other?

Microsoft.Azure.CosmosRepository/src/Builders/IItemContainerBuilder.cs
mumby0168: Will amend this must have missed it off.

mumby0168:
> I really like this! Looking great, but I'm curious what happens if you do not call the `.WithContainer` and `.WithPartitionKey` -- what is the behavior then? Also, what if you only call one or the other? Also, in terms of using the attributes and reflection - we are caching and it's really minimal. I do not believe that there is a major perf concern, yet. We could optimize this a bit by having a shared cache where the key is the `Type` and an object that holds all the details instead of maintaining multiple dictionaries.

Thanks for the review @IEvangelist,

Yeah so the way I went about this bit was almost a hierarchy it checks to see if an `Item` has been configure using the builder provided in the options, if so then it uses that value if that value is not defined (`null` or `""`) then it would check for an attribute if this exists then it will use the value from that if not as a last resort it will fall back to the name of the class for containers and `/id` for the partition key as a set of defaults.

In terms of the refactoring I think that does make sense having some sort of registry for options for a given type. Would you see this coming as part of this PR or as part of another issue and a future PR? 

I also have some more ideas around this where we can look at maybe providing a `ttl` option for items via this mechanism as well.

IEvangelist (dismissed):

dcuccia:
> Yeah so the way I went about this bit was almost a hierarchy it checks to see if an `Item` has been configure using the builder provided in the options, if so then it uses that value if that value is not defined (`null` or `""`) then it would check for an attribute if this exists then it will use the value from that if not as a last resort it will fall back to the name of the class for containers and `/id` for the partition key as a set of defaults.

I like your approach here. One additional thought is that we could also allow a strongly typed mapping configuration:
```csharp
options.ContainerBuilder.Configure<Product>(
    containerOptions => containerOptions
        .WithContainer("products")
        .WithPartitionKey(p => p.Category) // or, WithPartitionKeyPath(...)
);
```
...but then again, maybe less is more. One can always do `nameof(Product.Category).ToCamelCase()` if they want strong typing.

dcuccia (approved):

mumby0168:
@IEvangelist @dcuccia @bradwestness sorry guys, would you mind reviewing again? Had a merge to  sort and one comment to address. Then I can get this merged in.

IEvangelist (approved):

PR #82, opened 2021-08-06
Pagination with Continuation Token Support

### LINKED ISSUE #53: Feature request: pagination support

mumby0168:
WIP at the minute just to gather some feedback. 

Closes #53.

### DIFF
--- a/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
@@ -1,0 +1,108 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Bogus;
+using Microsoft.Azure.CosmosRepository;
+using Microsoft.Azure.CosmosRepository.Paging;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using static System.Console;
+
+ConfigurationBuilder configuration = new ConfigurationBuilder();
+
+ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
+{
+    options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+    options.DatabaseId = "paging-db";
+    options.ContainerPerItemType = true;
+})
+.AddSingleton<IConfiguration>(configuration.Build())
+.BuildServiceProvider();
+
+IRepository<Person> repository = provider.GetRequiredService<IRepository<Person>>();
+
+await PageWithContinuationTokens();
+await JumpStraightToPage();
+// await JumpToStraightThenUseContinuationToken();
+
+
+
+
+
+
+async Task JumpToStraightThenUseContinuationToken()
+{
+    int pageNumber = 4;
+    int pageSize = 10;
+    int totalItemsRead = 0;
+    int totalItemsExpectedToRead = 60;
+    string continuationToken = null;
+
+    double totalRuCharge = 0;
+
+    //TODO: this currently fails on the second pass when using the continuation token with a 400 bad request for a malformed token not entirely sure why yet.
+
+    while (totalItemsRead < totalItemsExpectedToRead)
+    {
+        IPage<Person> page = await repository.PageAsync(p => p.Age > 10, pageSize: pageSize, page: pageNumber, continuationToken: continuationToken);
+        continuationToken = page.ContinuationToken;
+        totalItemsRead += page.Items.Count();
+        pageNumber++;
+        totalRuCharge += page.Charge;
+        WriteLine($"Page {page.Number} returned {page.Items.Count()} results and cost {page.Charge} RUs");
+        WriteLine($"Names in page [{string.Join(",", page.Items.Select(i => $"{i.Name} ({i.Age})"))}]");
+        await Task.Delay(3000);
+    }
+
+    WriteLine($"Total charge for paging the whole container using continuation tokens {totalRuCharge} RUs");
+}
+
+
+async Task JumpStraightToPage()
+{
+    int pageNumber = 3;
+    int pageSize = 10;
+
+    IPage<Person> page = await repository.PageAsync(p => p.Age > 10, pageSize: pageSize, page: pageNumber);
+    pageNumber++;
+    WriteLine($"Page {page.Number} returned {page.Items.Count()} results and cost {page.Charge} RUs");
+    WriteLine($"Names in page [{string.Join(",", page.Items.Select(i => $"{i.Name} ({i.Age})"))}]");
+}
+
+
+
+async Task PageWithContinuationTokens()
+{
+    int pageNumber = 1;
+    int pageSize = 10;
+    int totalItemsRead = 0;
+    int totalItemsExpectedToRead = 100;
+    string continuationToken = null;
+
+    double totalRuCharge = 0;
+
+    while (totalItemsRead < totalItemsExpectedToRead)
+    {
+        IPage<Person> page = await repository.PageAsync(p => p.Age > 10, pageSize: pageSize, page: pageNumber, continuationToken: continuationToken);
+        continuationToken = page.ContinuationToken;
+        totalItemsRead += page.Items.Count();
+        pageNumber++;
+        totalRuCharge += page.Charge;
+        WriteLine($"Page {page.Number} returned {page.Items.Count()} results and cost {page.Charge} RUs");
+        WriteLine($"Names in page [{string.Join(",", page.Items.Select(i => $"{i.Name} ({i.Age})"))}]");
+    }
+
+    WriteLine($"Total charge for paging the whole container using continuation tokens {totalRuCharge} RUs");
+}
+
+async Task Seed()
+{
+    Faker<Person> peopleFaker = new();
+    peopleFaker
+        .RuleFor(p => p.Name, f => f.Name.FullName())
+        .RuleFor(p => p.Age, f => f.Random.Number(15, 45));
+
+    List<Person> people = peopleFaker.Generate(100);
+    await repository.CreateAsync(people);
+}
--- a/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
@@ -24,14 +24,14 @@ IRepository<Person> repository = provider.GetRequiredService<IRepository<Person>
 
 await PageWithContinuationTokens();
 await JumpStraightToPage();
-// await JumpToStraightThenUseContinuationToken();
+// await JumpToStraightToPage4ThenUseContinuationToken();
 
 
 
 
 
 
-async Task JumpToStraightThenUseContinuationToken()
+async Task JumpToStraightToPage4ThenUseContinuationToken()
 {
     int pageNumber = 4;
     int pageSize = 10;
--- a/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
@@ -21,15 +21,10 @@ ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =
 .BuildServiceProvider();
 
 IRepository<Person> repository = provider.GetRequiredService<IRepository<Person>>();
-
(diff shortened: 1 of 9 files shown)


Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    int pageNumber = 4;
+    int pageSize = 10;
+    int totalItemsRead = 0;
+    int totalItemsExpectedToRead = 60;
+    string continuationToken = null;
+
+    double totalRuCharge = 0;
+
+    //TODO: this currently fails on the second pass when using the continuation token with a 400 bad request for a malformed token not entirely sure why yet.
```
mumby0168, line 44: Having some trouble with this example the initial jump the page works correctly but then the subsequent request using a continuation token fails with a 400 bad request saying the token is malformed I will get a sample of the token and paste it here. 

I did look at this alternate implementation which did work but would hog memory & gets slower per page being iterated. I did not think this was practical https://gist.github.com/mumby0168/5553fd11dfac0cd4cd42c5a53ef121ae

mumby0168: This is the exception message

```
Unhandled exception. Microsoft.Azure.Cosmos.CosmosException : Response status code does not indicate success: BadRequest (400); Substatus: 0; ActivityId: ; Reason: (Response status code does not indicate success: BadRequest (400); Substatus: 0; ActivityId: ; Reason: (Response status code does not indicate su
ccess: BadRequest (400); Substatus: 0; ActivityId: ; Reason: (Response status code does not indicate success: BadRequest (400); Substatus: 0; ActivityId: ; Reason: (Invalid format for continuation token {"limit":2147483637,"sourceToken":"{\"offset\":0,\"sourceToken\":\"[{\\\"token\\\":\\\"+RID:~pR5MANaj6wwoA
AAAAAAAAA==#RT:4#TRC:40#ISV:2#IEO:65567#QCF:7#FPC:ASgAAAAAAAAAZAAAAAAAAAA=\\\",\\\"range\\\":{\\\"min\\\":\\\"\\\",\\\"max\\\":\\\"FF\\\"}}]\"}"} for ParallelCrossPartitionQueryPipelineStage););););
 ---> Microsoft.Azure.Cosmos.Query.Core.Exceptions.MalformedContinuationTokenException: Invalid format for continuation token {"limit":2147483637,"sourceToken":"{\"offset\":0,\"sourceToken\":\"[{\\\"token\\\":\\\"+RID:~pR5MANaj6wwoAAAAAAAAAA==#RT:4#TRC:40#ISV:2#IEO:65567#QCF:7#FPC:ASgAAAAAAAAAZAAAAAAAAAA=\\\
",\\\"range\\\":{\\\"min\\\":\\\"\\\",\\\"max\\\":\\\"FF\\\"}}]\"}"} for ParallelCrossPartitionQueryPipelineStage


```

mumby0168:
I have opened an issue to get some help on the issue we are facing at the minute: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2657

IEvangelist:
Hey, @zhangzunke I wanted to make sure that you saw this effort. Any thoughts on this implementation?

mumby0168:
@dcuccia @bradwestness Sorry guys mind taking a look at this?

joemersonsouza:
Hey @mumby0168 did you received some answer from Azure about the issue?

mumby0168:
> Hey @mumby0168 did you received some answer from Azure about the issue?

Hi no not yet, going to chase them, their is an issue open on GitHub & I have a ticket raised via my company. Will update on here when I hear back.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -37,6 +38,46 @@ internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : IItem
...
+            if (page < 1) throw new ArgumentOutOfRangeException(nameof(page));
+            if (pageSize < 1) throw new ArgumentOutOfRangeException(nameof(pageSize));
+
+
+            Container container = await _containerProvider.GetContainerAsync().ConfigureAwait(false);
+            Expression<Func<TItem, bool>> queryPredicate = predicate.Compose(item => !item.Type.IsDefined() || item.Type == typeof(TItem).Name, Expression.AndAlso);
+            IQueryable<TItem> query = null;
+            QueryRequestOptions queryOptions = new() { MaxItemCount = pageSize };
+
+            if (continuationToken is null)
```
dcuccia, line 53: Don't see any difference in the `continuationToken` branch. How about simplifying:

```csharp
query = container.GetItemLinqQueryable<TItem>(false, continuationToken, queryOptions).Where(queryPredicate);
if (page is not 1)
{
    query = query.Skip((page * pageSize) - pageSize);
}
```

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -31,6 +32,21 @@ namespace Microsoft.Azure.CosmosRepository
...
+        /// </summary>
+        /// <param name="predicate"></param>
+        /// <param name="cancellationToken"></param>
+        /// <param name="pageSize"></param>
+        /// <param name="page"></param>
+        /// <param name="continuationToken"></param>
+        /// <returns></returns>
+        ValueTask<IPage<TItem>> PageAsync(Expression<Func<TItem, bool>> predicate,
+            CancellationToken cancellationToken = default, int pageSize = 25, int page = 1,
+            string continuationToken = null);
```
dcuccia, line 47: Haven't read through the rest of the API, but should this be `string? continuationToken = null`?

Microsoft.Azure.CosmosRepository/src/Paging/Page.cs
```diff
@@ -0,0 +1,51 @@
...
+        public Page(int total, int number, int size, IEnumerable<T> items, double charge, string? continuationToken = null)
+        {
+            Total = total;
+            Number = number;
+            Size = size;
+            Items = items;
+            Charge = charge;
+            ContinuationToken = continuationToken;
+        }
+
```
dcuccia, line 31: Nit - extra line

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -37,6 +38,46 @@ internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : IItem
...
+            {
+                FeedResponse<TItem> next = await iterator.ReadNextAsync(cancellationToken);
+                items.AddRange(next.Resource);
+                charge += next.RequestCharge;
+                continuationToken = next.ContinuationToken;
+            }
+
+            return new Page<TItem>(total, page, pageSize, items, charge, continuationToken);
+        }
+
```
dcuccia, line 78: Nit - extra lines

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -37,6 +38,46 @@ internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : IItem
             ILogger<DefaultRepository<TItem>> logger) =>
             (_optionsMonitor, _containerProvider, _logger) = (optionsMonitor, containerProvider, logger);
 
+        public async ValueTask<IPage<TItem>> PageAsync(Expression<Func<TItem, bool>> predicate, CancellationToken cancellationToken = default, int pageSize = 25,
+            int page = 1, string continuationToken = null)
+        {
+            if (page < 1) throw new ArgumentOutOfRangeException(nameof(page));
+            if (pageSize < 1) throw new ArgumentOutOfRangeException(nameof(pageSize));
+
```
dcuccia, line 46: Nit - extra line

dcuccia (approved):
Approved with one suggestion for simplification. Excited to use this!

mumby0168:
Hey guys @IEvangelist @dcuccia @joemersonsouza so we have had a reply back from the cosmos team see https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2657#issuecomment-902897257

So it appears if you are to skip and get a continuation token then the next query must match exactly i.e. also including the original skip. We can incorporate this inside our library. It may add complications.

The 3 queries I wanted to support for paging aimed to meet the given scenarios.

CT -> (Contiuation Tokens)

* A user starts at page 1 and then cycles page 1,2,3 .... in order (this can make use of CT)
* A user wants to jump to a page say 4 and then cycle 5,6,7 in order again can use a CT but must know about the original skip.

Whether we support the second case is up for discussion? This would not support traditional paging but could be great for supporting a `load more ...` scenario.

I would really appreciate some feedback/help on what we think this should support. Obviously, CT provides a much more cost-effective way to page through data but also could potentially limit a user's experience. Definitely up for debate.

dcuccia:
@mumby0168 thanks for the follow-up. Could we support both scenarios, but somehow unify the experience by returning an `IAsyncEnumerable<IPage>`? Or, if that's the API, then does it mean we can remove the second one altogether, since the caller will have control over the "load more" iteration? Armchair development, so caveat emptor...

mumby0168:
> @mumby0168 thanks for the follow-up. Could we support both scenarios, but somehow unify the experience by returning an `IAsyncEnumerable<IPage>`? Or, if that's the API, then does it mean we can remove the second one altogether, since the caller will have control over the "load more" iteration? Armchair development, so caveat emptor...

Yeah maybe so, it depends how we want to support it, I have been coming at this from the point of view that a user is on a screen look at a blog. They are on the first page and want to view the next. This would then come in two separate requests more than likely an API. I thought about `IAsyncEnumerable<T>` but would that work well for a scenario such as above? Not really worked with it too often so that is a limited view. 

I suppose we may just have to rely on the caller knowing a bit about what is going on, matching a skip up would not be to hard, but then if this is on top of an `OrderBy(x => ...)` for example the problem grows 🤔

mumby0168:
Part of me is thinking about closing this PR for now. Maybe opening up a new discussion section around the design of this. But at the same time this is something I would love to see soon 👀

dcuccia:
My thought is that if we expose the async iterator, then we can encapsulate the complexity and keep the CT knowledge behind the scenes. Leaving open for more comments/thoughts is fine with me. I may have a little time to play with the iterator concept.

mumby0168:
> My thought is that if we expose the async iterator, then we can encapsulate the complexity and keep the CT knowledge behind the scenes. Leaving open for more comments/thoughts is fine with me. I may have a little time to play with the iterator concept.

Yeah that would be great to see! No reason we could not offer two options ... is there?

dcuccia:
@mumby0168 I think we might be able to offer both via a single API. Check out what I just pushed - does this cover our needs? Using the `.FirstAsync()` extension for the case where you "just want one page", but that would usually be the exception. The requirement for that case is that the consumer pull in System.Linq.Async, which is currently not shipped with the SDK (but I imagine it will be soon).

dcuccia:
@mumby0168 reading back, I finally understand what you mean about the separate-call API scenario. I don't know how to resolve that one as elegantly...maybe this is the minimum viable approach? Not worth it if it doesn't solve a major use case, though.

dcuccia:
Maybe there's a state machine we can expose on *top* of the IAsyncEnumerable. Our own continuation async iterator...

mumby0168:
Hey @dcuccia I think the push you made looks really cool and certainly would be good to add as an extra option to the API! I think this paging issue might need more discussion what does everyone think about putting something in the calendar to go and discuss this and decide our approaches? @IEvangelist @bradwestness and anyone else who would like to discuss?

mumby0168:
@IEvangelist what about offering a continuation token paging option, an `IAsyncEnumerable` option like @dcuccia suggested and then just expose the `Container` via the `IRepository<T>` if a client wants to handle this them selves? Really struggling to decide here 🤦‍♂️

mumby0168:
> Hey guys, do you have some news about this feature?

Hi @joemersonsouza we are in two minds with it, what would you like to see and what are your requirements for paging if you don't mind?

mumby0168:
I am going to close this PR it is worth refereeing back to I think but we have some discussions set up on how to go about this further in the next few weeks. Hopefully, after that, we can have a plan of attack on how to implement this correctly. 

Thanks, everyone for the discussions/contributions to this so far.

PR #87, opened 2021-08-13
Auto deploy to nuget on a tag

mumby0168:
This will start the build when a tag is added to master branch such as `v1.1` this then extracts the version number from it and versions the package with this version which is then published to nuget.org.

### DIFF
--- a/.github/workflows/nuget-publish.yml
+++ b/.github/workflows/nuget-publish.yml
@@ -41,7 +41,7 @@ jobs:
       
     - name: Pack
       run: |
-        dotnet pack ./Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj -p:PackageVersion='${{ steps.regex-match.outputs.match }}' --output packages
+        dotnet pack ./Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj --output packages
       
     - name: Publish Package
       run: |
--- a/.github/workflows/nuget-publish.yml
+++ b/.github/workflows/nuget-publish.yml
@@ -1,0 +1,55 @@
+name: Release Package Version
+on:
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/nuget-publish.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: |
+        dotnet build --configuration Release --no-restore
+      
+    - name: Test
+      run: |
+        dotnet test --no-restore --verbosity normal
+      
+    - name: Pack
+      run: |
+        dotnet pack ./Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj -p:PackageVersion='${{ steps.regex-match.outputs.match }}' --output packages
```
IEvangelist, line 44:
```suggestion
        dotnet pack ./Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj --output packages
```

You shouldn't need to manually apply the version, this is handled implicitly on your behalf from MinVer based on the tag. See [here](https://rehansaeed.com/the-easiest-way-to-version-nuget-packages/).

IEvangelist (commented):
One nit, other than that - this looks great!

.github/workflows/nuget-publish.yml
mumby0168: Ahh okay I see that is pretty cool.

mumby0168:
> One nit, other than that - this looks great!

Yeah did not realise that `MinVer` was a nuget package, this will be really cool, would be good to get some of the recent changes released as well 👍

IEvangelist (approved):

mumby0168:
@dcuccia @bradwestness Sorry, anyone able to have a quick look at this please?

bradwestness (approved):
Yaml-icious!

PR #93, opened 2021-08-20
Aggregated configuration into a single location

mumby0168:
Just some general tidy up we had 2 or 3 classes all using a `ConcurrentDictionary` thought it would be better served to just have the one.

This is to pave the way for extra configuration hopefully make it easier to build on it.


Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosItemConfigurationProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        public DefaultCosmosItemConfigurationProvider(
+            ICosmosContainerNameProvider containerNameProvider,
+            ICosmosPartitionKeyPathProvider cosmosPartitionKeyPathProvider,
+            ICosmosUniqueKeyPolicyProvider cosmosUniqueKeyPolicyProvider
+        )
+        {
+            _containerNameProvider = containerNameProvider;
+            _cosmosPartitionKeyPathProvider = cosmosPartitionKeyPathProvider;
+            _cosmosUniqueKeyPolicyProvider = cosmosUniqueKeyPolicyProvider;
+        }
```
IEvangelist, line 28:
```suggestion
            ICosmosUniqueKeyPolicyProvider cosmosUniqueKeyPolicyProvider)
        {
            _containerNameProvider = containerNameProvider;
            _cosmosPartitionKeyPathProvider = cosmosPartitionKeyPathProvider;
            _cosmosUniqueKeyPolicyProvider = cosmosUniqueKeyPolicyProvider;
        }
```

Inconsistent styling.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosItemConfigurationProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public ItemOptions GetOptions<TItem>() where TItem : IItem
+            => ItemOptionsMap.GetOrAdd(typeof(TItem), AddOptions<TItem>);
```
IEvangelist, line 31:
```suggestion
        public ItemOptions GetOptions<TItem>() where TItem : IItem =>
            ItemOptionsMap.GetOrAdd(typeof(TItem), AddOptions<TItem>);
```

Prefer goes to operator on the line prior to the corresponding expression.

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosItemConfigurationProviderTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using Xunit;
+
+namespace Microsoft.Azure.CosmosRepositoryTests.Providers
+{
+    public class DefaultCosmosItemConfigurationProviderTests
+    {
+        private readonly Mock<ICosmosContainerNameProvider> _containerNameProvider = new();
+        private readonly Mock<ICosmosPartitionKeyPathProvider> _partitionKeyPathProvider = new();
+        private readonly Mock<ICosmosUniqueKeyPolicyProvider> _uniqueKeyPolicyProvider = new();
+
```
IEvangelist, line 18:
```suggestion
```

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosItemConfigurationProviderTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        [Fact]
+        public void GetOptionsAlwaysGetOptionsForItem()
+        {
+            ICosmosItemConfigurationProvider provider = new DefaultCosmosItemConfigurationProvider(
+                _containerNameProvider.Object,
+                _partitionKeyPathProvider.Object,
+                _uniqueKeyPolicyProvider.Object
+            );
```
IEvangelist, line 27:
```suggestion
                _uniqueKeyPolicyProvider.Object);
```

Inconsistent styling.

IEvangelist (commented):
A few nits, curious about this... Is this a breaking change?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosItemConfigurationProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using System.Collections.Concurrent;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Azure.CosmosRepository.Options;
+
+namespace Microsoft.Azure.CosmosRepository.Providers
+{
+    class DefaultCosmosItemConfigurationProvider : ICosmosItemConfigurationProvider
+    {
+        private static readonly ConcurrentDictionary<Type, ItemOptions> ItemOptionsMap = new();
```
IEvangelist, line 13:
```suggestion
        private static readonly ConcurrentDictionary<Type, ItemOptions> s_itemOptionsMap = new();
```

Inconsistent styling, static fields should be `s_` prefixed.

IEvangelist:
Ok, I see what you're doing here and I like it. I had thought about doing something similar! So, we need to ensure that the `ItemOptions` is non-destructive. For example, for a given type each `ItemOptions` in the dictionary needs to be additive. What you have now is essentially last config setting wins and it will lose previously set values. Perhaps `ItemOptions` could be a record, and if there is already a value for a given type key - you could use the `with` syntax to clone the existing, and update with the new values.

mumby0168:
> A few nits, curious about this... Is this a breaking change?

Yeah no worries will sort those out soon, no wasn't intended to be, just the way the configuration is stored. The public API hasn't changed.

IEvangelist:
> > A few nits, curious about this... Is this a breaking change?
> 
> Yeah no worries will sort those out soon, no wasn't intended to be, just the way the configuration is stored. The public API hasn't changed.

A public type was renamed, if anyone is consuming this and had an instance of that type with explicit typing when declaring the variable - this would break them, right? 😄

mumby0168:
> Hey @mumby0168 
> 
> 
> 
> Ok, I see what you're doing here and I like it. I had thought about doing something similar! So, we need to ensure that the `ItemOptions` is non-destructive. For example, for a given type each `ItemOptions` in the dictionary needs to be additive. What you have now is essentially last config setting wins and it will lose previously set values. Perhaps `ItemOptions` could be a record, and if there is already a value for a given type key - you could use the `with` syntax to clone the existing, and update with the new values.

Hey, forgive me if I have misunderstood but once the options have been added to the dictionary, they are almost locked in at that point with this been a static dictionary on a singleton object once they have been added to the dictionary they will be there for good? Is what you are saying implying that the values configured here could change on the fly ?

mumby0168:
> > > A few nits, curious about this... Is this a breaking change?
> > 
> > 
> > Yeah no worries will sort those out soon, no wasn't intended to be, just the way the configuration is stored. The public API hasn't changed.
> 
> A public type was renamed, if anyone is consuming this and had an instance of that type with explicit typing when declaring the variable - this would break them, right? 😄 It's not a big deal, was just calling attention to it.

Ahh yes the `ContainerOptions` do you think I should change it back? That only went out with the last release?

IEvangelist:
> > > > A few nits, curious about this... Is this a breaking change?
> > > 
> > > 
> > > Yeah no worries will sort those out soon, no wasn't intended to be, just the way the configuration is stored. The public API hasn't changed.
> > 
> > 
> > A public type was renamed, if anyone is consuming this and had an instance of that type with explicit typing when declaring the variable - this would break them, right? 😄 It's not a big deal, was just calling attention to it.
> 
> Ahh yes the `ContainerOptions` do you think I should change it back? That only went out with the last release?

No. It's probably not a big deal.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosItemConfigurationProvider.cs
mumby0168: Amended.

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosItemConfigurationProviderTests.cs
mumby0168: Amended.

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosItemConfigurationProviderTests.cs
mumby0168: Amended.

IEvangelist:
> > Hey @mumby0168
> > Ok, I see what you're doing here and I like it. I had thought about doing something similar! So, we need to ensure that the `ItemOptions` is non-destructive. For example, for a given type each `ItemOptions` in the dictionary needs to be additive. What you have now is essentially last config setting wins and it will lose previously set values. Perhaps `ItemOptions` could be a record, and if there is already a value for a given type key - you could use the `with` syntax to clone the existing, and update with the new values.
> 
> Hey, forgive me if I have misunderstood but once the options have been added to the dictionary, they are almost locked in at that point with this been a static dictionary on a singleton object once they have been added to the dictionary they will be there for good? Is what you are saying implying that the values configured here could change on the fly ?

I might be missing something too, from what it looks like 

> > Hey @mumby0168
> > Ok, I see what you're doing here and I like it. I had thought about doing something similar! So, we need to ensure that the `ItemOptions` is non-destructive. For example, for a given type each `ItemOptions` in the dictionary needs to be additive. What you have now is essentially last config setting wins and it will lose previously set values. Perhaps `ItemOptions` could be a record, and if there is already a value for a given type key - you could use the `with` syntax to clone the existing, and update with the new values.
> 
> Hey, forgive me if I have misunderstood but once the options have been added to the dictionary, they are almost locked in at that point with this been a static dictionary on a singleton object once they have been added to the dictionary they will be there for good? Is what you are saying implying that the values configured here could change on the fly ?

Disregard, I was missing that that the `DefaultCosmosItemConfigurationProvider` aggregates all the attribute based discovery. This is beautiful!

IEvangelist (approved):

bradwestness (approved):
Rad!

PR #110, opened 2021-09-12
Added default time to live to a container configuration

### LINKED ISSUE #80: Set TTL (Time to Live) at Container Level
`n/a`

A way to specify the default time to live for a whole container. This can be achieved via the code below. It would be a good feature to have this configurable for each container
```csharp
CosmosClient cosmosClient = new CosmosClient("<conn-string>");
Database database = cosmosClient.GetDatabase("<database-id>");
await database.CreateContainerIfNotExistsAsync(new ContainerProperties
{
    DefaultTimeToLive = TimeSpan.FromDays(5).Seconds
});
```

`n/a`

**Additional context**
`n/a`

mumby0168:
Closes #80.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
@@ -16,23 +16,40 @@ namespace Microsoft.Azure.CosmosRepository.Builders
         internal Type Type { get; }
 
         /// <summary>
-        /// Creates an instance of <see cref="ContainerOptionsBuilder"/>
+        /// Creates an instance of <see cref="ContainerOptionsBuilder"/>.
         /// </summary>
-        /// <param name="type">The type of <see cref="IItem"/> the options are for</param>
+        /// <param name="type">The type of <see cref="IItem"/> the options are for.</param>
         public ContainerOptionsBuilder(Type type) => Type = type;
 
         /// <summary>
-        /// Name of the container
+        /// Name of the container.
         /// </summary>
         internal string Name { get; private set; }
 
         /// <summary>
-        /// The partition key for the container
+        /// The partition key for the container.
         /// </summary>
         internal string PartitionKey { get; private set; }
 
         /// <summary>
-        /// Sets the name of the container
+        /// The default time to live for a container.
+        /// </summary>
+        /// <remarks>If <see cref="Item"/> share a container they will share this property.</remarks>
+        internal TimeSpan? ContainerDefaultTimeToLive { get; private set; }
+
+        /// <summary>
+        /// Sets the <see cref="ContainerDefaultTimeToLive"/> for a container.
+        /// </summary>
+        /// <param name="containerDefaultTimeToLive">The default time to live for the container.</param>
+        /// <returns></returns>
+        public ContainerOptionsBuilder WithContainerDefaultTimeToLive(TimeSpan containerDefaultTimeToLive)
+        {
+            ContainerDefaultTimeToLive = containerDefaultTimeToLive;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the <see cref="Name"/> of the container
         /// </summary>
         /// <param name="name">The name of the container</param>
         /// <returns>Instance of <see cref="ContainerOptionsBuilder"/></returns>
--- a/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.CosmosRepository.Providers
 {
 
     /// <inheritdoc/>
-    public class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
+    class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
     {
         private readonly IOptions<RepositoryOptions> _options;
 
--- a/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
@@ -1,0 +1,47 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+using Microsoft.Azure.CosmosRepository.Builders;
+using Microsoft.Azure.CosmosRepository.Options;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.CosmosRepository.Providers
+{
+
+    /// <inheritdoc/>
+    public class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
+    {
+        private readonly IOptions<RepositoryOptions> _options;
+
+        /// <summary>
+        /// Creates an instance of the <see cref="DefaultCosmosContainerDefaultTimeToLiveProvider"/>.
+        /// </summary>
+        /// <param name="options">The repository options.</param>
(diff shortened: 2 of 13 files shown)


mumby0168:
Had a thought while work on this issue. We currently are working towards offering more configuration for containers including the work for #80. What occurs if a client has a container that is created, then they say updates the default time to live for that container? I imagine currently as we are using the `CreateContainerIfNotExistsAsync` method that it would not work as expected if we were to update the container's time to live.

I was thinking is it worth adding something that can be run on startup that will amend any container's properties if they are not the same as what is already defined? I do need to verify that modifying a container via the SDK is possible but I do think it would be worth looking at as we add more support for container configuration.

Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
```diff
@@ -16,23 +16,40 @@ public class ContainerOptionsBuilder
...
+        /// The default time to live for a container.
+        /// </summary>
+        /// <remarks>If <see cref="Item"/> share a container they will share this property.</remarks>
+        internal TimeSpan? ContainerDefaultTimeToLive { get; private set; }
+
+        /// <summary>
+        /// Sets the <see cref="ContainerDefaultTimeToLive"/> for a container.
+        /// </summary>
+        /// <param name="containerDefaultTimeToLive">The default time to live for the container.</param>
+        /// <returns></returns>
```
IEvangelist, line 44:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.Azure.CosmosRepository.Builders;
+using Microsoft.Azure.CosmosRepository.Options;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.CosmosRepository.Providers
+{
+
```
IEvangelist, line 12:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <inheritdoc/>
+    public class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
```
IEvangelist, line 14: I'm curious, are our other default providers `public`? I can't recall if we hid them as `internal`, not to expose the implementation as that is a detail the consumer doesn't need to know.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// </summary>
+        /// <param name="options">The repository options.</param>
+        public DefaultCosmosContainerDefaultTimeToLiveProvider(IOptions<RepositoryOptions> options) => _options = options;
+
+        /// <inheritdoc/>
+        public int GetDefaultTimeToLive<TItem>() where TItem : IItem
+        {
+            ContainerOptionsBuilder options = _options.Value.ContainerOptions.FirstOrDefault(co => co.Type == typeof(TItem));
+
+            if (options is null)
```
IEvangelist, line 29:
```suggestion
            var options = _options.Value.ContainerOptions?.FirstOrDefault(co => co.Type == typeof(TItem));
            if (options is null)
```

Better safe, than sorry. I'm not sure if this can be `null` or not?

Microsoft.Azure.CosmosRepository/test/Providers/DefaultCosmosContainerDefaultTimeToLiveProviderTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            Assert.Equal(-1, _provider.GetDefaultTimeToLive<TestItem>());
+
+        [Fact]
+        public void GetDefaultTimeToLiveItemWithOptionsUsesThatTimeToLive()
+        {
+            TimeSpan timeToLive = TimeSpan.FromMinutes(5);
+            _repositoryOptions.ContainerBuilder.Configure<TestItem>(builder =>
+                builder.WithContainerDefaultTimeToLive(timeToLive));
+
+
```
IEvangelist, line 30:
```suggestion
```

IEvangelist (approved):
Approved, but conditional on my suggestions and questions.

IEvangelist:
> Had a thought while work on this issue. We currently are working towards offering more configuration for containers including the work for #80. What occurs if a client has a container that is created, then they say updates the default time to live for that container? I imagine currently as we are using the `CreateContainerIfNotExistsAsync` method that it would not work as expected if we were to update the container's time to live.
> 
> I was thinking is it worth adding something that can be run on startup that will amend any container's properties if they are not the same as what is already defined? I do need to verify that modifying a container via the SDK is possible but I do think it would be worth looking at as we add more support for container configuration.

That's an interesting thought. I'm curious:

- how often that kind of thing would occur
- how do we run it asynchronously at startup
- do the benefits outweigh the dev effort
- does this introduce issues with backward compat

mumby0168:
> That's an interesting thought. I'm curious:
> 
> * how often that kind of thing would occur
> * how do we run it asynchronously at startup
> * do the benefits outweigh the dev effort
> * does this introduce issues with backward compact

In the answer to those questions.

1. we are seeing it at the minute at my company as there is a push for infrastructure as code and it could occur if let's say an indexing policy updates or the time to live would change say customer's order want to remain for 4 months now instead of 3?
2. I propose we don't run this at startup but maybe (opt-in only of course) run this as part of the `Lazy` load of the container when the repository instance is first created.
3. Personally I think this would be a really cool push to the infrastructure of code and allowing the application to control the state of the containers.
4. I am not sure it would if we make this opt-in only. I would certainly not want people who didn't want to pay the cost of the first request to check to have to do this check/update of the container.

That all been said we can offer a manual option via another service exposed via the library to sync up container settings, which would then push the responsibility of the async startup to the user. If a user did want to do it at start-up but opt-out of it loading up at the point that the first instance container is initialized (opting out of point 2.). 

We can create an issue and tag these comments as part of it for further discussion what do you think?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
mumby0168: You are correct this should be internal.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
mumby0168: Container options are initialized in the class with `new()` so it can never be null.

IEvangelist (approved):

mumby0168:
> > That's an interesting thought. I'm curious:
> > 
> > * how often that kind of thing would occur
> > * how do we run it asynchronously at startup
> > * do the benefits outweigh the dev effort
> > * does this introduce issues with backward compact
> 
> In the answer to those questions.
> 
> 1. we are seeing it at the minute at my company as there is a push for infrastructure as code and it could occur if let's say an indexing policy updates or the time to live would change say customer's order want to remain for 4 months now instead of 3?
> 2. I propose we don't run this at startup but maybe (opt-in only of course) run this as part of the `Lazy` load of the container when the repository instance is first created.
> 3. Personally I think this would be a really cool push to the infrastructure of code and allowing the application to control the state of the containers.
> 4. I am not sure it would if we make this opt-in only. I would certainly not want people who didn't want to pay the cost of the first request to check to have to do this check/update of the container.
> 
> That all been said we can offer a manual option via another service exposed via the library to sync up container settings, which would then push the responsibility of the async startup to the user. If a user did want to do it at start-up but opt-out of it loading up at the point that the first instance container is initialized (opting out of point 2.).
> 
> We can create an issue and tag these comments as part of it for further discussion what do you think?

More discussion can be found here #111.

PR #111, opened 2021-09-13
Container syncing

### LINKED ISSUE #112: Container Syncing
The new configuration options that have been provided as part of this library may mean that containers become out of sync with what is configured within this library.

An opt-in option to sync container properties when the container is initially lazy-loaded can allow clients to keep their container settings in sync.

There should also be a manual option provided via a DI-enabled service to manually sync these containers if clients don't want to take the hit on the first request for that container.

mumby0168:
Closes #112

### DIFF
--- a/Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Startup.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Startup.cs
@@ -17,7 +17,7 @@ namespace AzureFunctionTier
                     .WithContainer("users")
                     .WithPartitionKey("/emailAddress")
                     .WithContainerDefaultTimeToLive(TimeSpan.FromMinutes(1))
-                    .WithSyncedContainerProperties()
+                    .WithSyncableContainerProperties()
                 );
             });
     }
--- a/Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Startup.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Startup.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.Azure.Cosmos;
+﻿using System;
+using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.Functions.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection;
 using User = AzureFunctionTier.Model.User;
@@ -15,6 +16,8 @@ namespace AzureFunctionTier
                 options.ContainerBuilder.Configure<User>(containerOptions => containerOptions
                     .WithContainer("users")
                     .WithPartitionKey("/emailAddress")
+                    .WithContainerDefaultTimeToLive(TimeSpan.FromMinutes(1))
+                    .WithSyncedContainerProperties()
                 );
             });
     }
--- a/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.CosmosRepository.Providers
         /// <inheritdoc/>
         public int GetDefaultTimeToLive<TItem>() where TItem : IItem
         {
-            ContainerOptionsBuilder options = _options.Value.ContainerOptions.FirstOrDefault(co => co.Type == typeof(TItem));
+            ContainerOptionsBuilder options = _options.Value.GetContainerOptions<TItem>();
 
             if (options is null)
             {
--- a/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.CosmosRepository.Providers
 {
 
     /// <inheritdoc/>
-    public class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
+    class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
     {
         private readonly IOptions<RepositoryOptions> _options;
 
--- a/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
@@ -1,0 +1,47 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+using Microsoft.Azure.CosmosRepository.Builders;
+using Microsoft.Azure.CosmosRepository.Options;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.CosmosRepository.Providers
+{
+
+    /// <inheritdoc/>
+    public class DefaultCosmosContainerDefaultTimeToLiveProvider : ICosmosContainerDefaultTimeToLiveProvider
+    {
+        private readonly IOptions<RepositoryOptions> _options;
+
+        /// <summary>
+        /// Creates an instance of the <see cref="DefaultCosmosContainerDefaultTimeToLiveProvider"/>.
+        /// </summary>
+        /// <param name="options">The repository options.</param>
+        public DefaultCosmosContainerDefaultTimeToLiveProvider(IOptions<RepositoryOptions> options) => _options = options;
+
+        /// <inheritdoc/>
+        public int GetDefaultTimeToLive<TItem>() where TItem : IItem
+        {
+            ContainerOptionsBuilder options = _options.Value.ContainerOptions.FirstOrDefault(co => co.Type == typeof(TItem));
+
+            if (options is null)
+            {
+                return -1;
+            }
+
+            foreach (ContainerOptionsBuilder containerOptions in _options.Value.ContainerOptions.Where(co => co.Name == options.Name))
+            {
+                if (containerOptions.ContainerDefaultTimeToLive != null && containerOptions.ContainerDefaultTimeToLive != options.ContainerDefaultTimeToLive)
+                {
+                    throw new InvalidOperationException($"The container {options.Name} has conflicting default time to live values. " +
+                                                        $"({options.Type.Name}->{options.ContainerDefaultTimeToLive} vs " +
+                                                        $"{containerOptions.Type.Name}->{containerOptions.ContainerDefaultTimeToLive}).");
+                }
+            }
+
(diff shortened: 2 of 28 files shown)


Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (options is null)
+            {
+                return -1;
+            }
+
+            foreach (ContainerOptionsBuilder containerOptions in _options.Value.ContainerOptions.Where(co => co.Name == options.Name))
+            {
+                if (containerOptions.ContainerDefaultTimeToLive != null && containerOptions.ContainerDefaultTimeToLive != options.ContainerDefaultTimeToLive)
+                {
+                    throw new InvalidOperationException($"The container {options.Name} has conflicting default time to live values. " +
```
IEvangelist, line 37: Maybe something we could do in the future to evolve this library a bit more is document and promote official error codes and messages in a look up for consumers? With troubleshooting help for common difficulties?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
```diff
@@ -58,13 +58,19 @@ async Task<Container> GetContainerValueFactoryAsync()
...
-                    UniqueKeyPolicy = itemOptions.UniqueKeyPolicy ?? new()
+                    UniqueKeyPolicy = itemOptions.UniqueKeyPolicy ?? new(),
+                    DefaultTimeToLive = itemOptions.DefaultTimeToLive
                 };
 
                 Container container =
                     await database.CreateContainerIfNotExistsAsync(
                         containerProperties).ConfigureAwait(false);
 
+                if (itemOptions.SyncContainerProperties)
```
IEvangelist, line 69: So this bit has to be flipped with each deployed to avoid the call to `ReplaceContainerAsync`? Is this call idempotent, or like a noop? I'm curious about the overhead of this, if it is not truly needed.

IEvangelist (dismissed):

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
mumby0168: Yes I agree, how would you do that a custom exception from our library?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
mumby0168: Do you mean if two `IItem`s share a container then we do not need to replace the container more than once?

IEvangelist: Yes, if the app starts and does a replacement of the container - an expensive network I/O call. Then the app shuts down happily, and the next time it starts does it need to again run this same code? Or is it good now? My concern is that when this is set to `true` we're always adding I/O overhead to the startup time, which may not be necessary. I'm not sure if that is the case or not.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
IEvangelist: We could certainly start by creating exceptions, but I was thinking we could either build out a Wiki here too or even a stand alone (but that's a bigger effort). Or maybe not, I could probably spin up a Hugo doc site in an afternoon.

mumby0168: Yeah that would be really cool, could that be hosted on Github pages?

IEvangelist: Indeed, that could.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
mumby0168: Yes so if a client opts in to this then every time the ` Lazy<Task<Container>>` is resolved for the first time then yes it would make this `I/O` call. I have also added a service for this so a client can manually replace the container if needs be so they get to choose whether to pay this price for the I/O call or not.

Suppose with the service they could do this as part of a health check or something?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
mumby0168: That would be great, is it markdown based?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
mumby0168: This happens on the first request that involves a request to cosmos, not to the startup time of the application.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
IEvangelist: Yes, entirely markdown based. I'll see what I can do, but time is something I don't have a lot of lately. :)

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
IEvangelist: If anything we just need to make that clear to consumers, so that they fully understand the potential ramifications of opting into it. It's almost like an EF migration.

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerDefaultTimeToLiveProvider.cs
mumby0168: I can always take a look if we get the repo up and working?

Microsoft.Azure.CosmosRepository/src/Providers/DefaultCosmosContainerProvider.cs
mumby0168: Yes, I agree it opted out by default so they would have to explicitly turn it on. Some docs on this would be good maybe even on the readme.

IEvangelist (dismissed):

Microsoft.Azure.CosmosRepository.Samples/AzureFunctionTier/Startup.cs
```diff
@@ -15,6 +16,8 @@ class Startup : FunctionsStartup
                 options.ContainerBuilder.Configure<User>(containerOptions => containerOptions
                     .WithContainer("users")
                     .WithPartitionKey("/emailAddress")
+                    .WithContainerDefaultTimeToLive(TimeSpan.FromMinutes(1))
+                    .WithSyncedContainerProperties()
```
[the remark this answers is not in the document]
mumby0168: Yes I think I agree I like this one `WithSyncableContainerProperties` will update.

IEvangelist (approved):

Microsoft.Azure.CosmosRepository/src/Services/DefaultCosmosContainerSyncService.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.Azure.CosmosRepository.Services
+{
+    /// <inheritdoc/>
+    class DefaultCosmosContainerSyncService : ICosmosContainerSyncService
+    {
+        readonly ICosmosContainerService _containerService;
+
+        public DefaultCosmosContainerSyncService(ICosmosContainerService containerService) => _containerService = containerService;
+
+        public Task SyncAsync<TItem>() where TItem : IItem => _containerService.GetContainerAsync<TItem>(true);
```
IEvangelist, line 18: I guess one more suggestion, do you like this API? It's rather vague. What about `SynchronizeContainerPropertiesAsync` or `SyncContainerPropertiesAsync`?

Microsoft.Azure.CosmosRepository/src/Services/ICosmosContainerSyncService.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.CosmosRepository.Services
+{
+    /// <summary>
+    /// Allows containers properties to be sunk.
+    /// </summary>
+    public interface ICosmosContainerSyncService
+    {
+
```
IEvangelist, line 13:
```suggestion
```

Microsoft.Azure.CosmosRepository/test/Providers/DefaultContainerSyncContainerPropertiesProviderTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            Assert.True(_provider.GetWhetherToSyncContainerProperties<TestItem>());
+        }
+
+        [Fact]
+        public void GetWhetherToSyncContainerPropertiesWhenOptionsToSyncIsFalse()
+        {
+            _repositoryOptions.ContainerBuilder.Configure<TestItem>(builder => { });
+            Assert.False(_provider.GetWhetherToSyncContainerProperties<TestItem>());
+        }
+
```
IEvangelist, line 42:
```suggestion
```

Microsoft.Azure.CosmosRepository/test/Services/DefaultCosmosContainerSyncServiceTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Azure.CosmosRepository.Services;
+using Microsoft.Azure.CosmosRepositoryTests.Stubs;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.CosmosRepositoryTests.Services
+{
+    public class DefaultCosmosContainerSyncServiceTests
+    {
+
```
IEvangelist, line 14:
```suggestion
```

Microsoft.Azure.CosmosRepository/test/Stubs/AnotherTestItem.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosRepository;
+
+namespace Microsoft.Azure.CosmosRepositoryTests.Stubs
+{
+    public class AnotherTestItem : Item
+    {
+
```
IEvangelist, line 10:
```suggestion
```

IEvangelist (dismissed):
It looks like a few more files slipped in, added a few more suggestions.

Microsoft.Azure.CosmosRepository/src/Services/DefaultCosmosContainerSyncService.cs
mumby0168: Yes, I think that is a fair comment, will amend this.

IEvangelist (approved):

PR #114, opened 2021-09-15
Added support to provide a managed identity option

### LINKED ISSUE #56: Feature : Managed identity
Thanks for the project!!
We want to access our database through managed Identity, when will it be possible ?

mumby0168:
Closes #56.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using Azure.Core;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.CosmosRepository.Builders;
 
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
```diff
@@ -80,6 +81,11 @@ public class RepositoryOptions
         /// </summary>
         public RepositorySerializationOptions SerializationOptions { get; set; }
 
+        /// <summary>
+        /// The <see cref="TokenCredential"/> which can be used to access azure resources, including Cosmos DB.
+        /// </summary>
+        public TokenCredential TokenCredential { get; set; } = null;
```
IEvangelist, line 87: Is this properly deserialized if set in the JSON configuration provider? In other words, does this work?
```json
{
    "RepositoryOptions": {
        "TokenCredential": {
            "Token": "51AE5DDxxxxxxxxxxxxxxxx556E2E"
        }
    }
}
```

Also, @mumby0168 I though of you when I was just looking at the API ref docs on the `CosmosClient` type. They have a builder pattern object for the client itself 🤓:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.fluent.cosmosclientbuilder?view=azure-dotnet

IEvangelist (approved):
Looks good to me, :shipit: when you're ready. But I did have one verification question.

Microsoft.Azure.CosmosRepository/src/Options/RepositoryOptions.cs
mumby0168: In the response to the first question, I am really unsure, would have to try it out, I know the pattern people usually follow is configuring this identity by using most of the time the `new DefaultAzureCredential();`. So would be cool if it did work but not sure what use case it would have?

No way we should totally add this builder to one of the samples!

PR #124, opened 2021-10-30
Only sync a shared container once

### LINKED ISSUE #122: Smarter Container Syncing
There is currently a mechanism in the library that will allow you to sync container properties on the first request to an `IRepository<TItem>` for a given `IItem` type. 

This currently will check the configured `ContainerProperties` by the library and then it will apply these to the cosmos container that the `IItem` is using. This does not take into account that `IItem`s are sharing a container potentially applying these changes for every `IItem` when the first request is made to the container for the given `IItem` type.

This could be more intelligent and keep track of sync containers and if a request to one `IItem`s repository has already been made which results in the container changes being sync then it skips this step.

The main issue with this is that the initial request takes time and reducing the times the `ContainerProperties` are synced for `IItem` types that share the same container will result in a performance increase.

mumby0168:
Closes #122

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Services/DefaultCosmosContainerService.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Services/DefaultCosmosContainerService.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


IEvangelist (approved):

PR #132, opened 2021-11-06
Support Partial Replace Updates

### LINKED ISSUE #128: Partial Document Update -  Replace
Support for the new general availability feature of a partial update would be really good to add see the announcement below

https://devblogs.microsoft.com/cosmosdb/partial-document-update-ga/

First thought's after looking at the .net code samples would be to have something such as 

```csharp
string id = "myid"
string pk = "myPartitionKey"
string newValue = "some-new-value"

_repository
.UpdateAsync(id, pk, item => item
.Replace(item => item.Property, newValue)
.Replace(item => item.OtherProperty, "newValue"));
```

# Raw API Examples

```csharp
List<PatchOperation> patchOperations = new List<PatchOperation>();
patchOperations.Add(PatchOperation.Add("/nonExistentParent/Child", "bar"));
patchOperations.Add(PatchOperation.Remove("/cost"));
patchOperations.Add(PatchOperation.Increment("/taskNum", 6));
patchOperations.Add(patchOperation.Set("/existingPath/newproperty",value));

container.PatchItemAsync<item>(
                id: 5,
                partitionKey: new PartitionKey("task6"),
                patchOperations: patchOperations );
```

mumby0168:
Closes #128

### DIFF
--- a/Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
@@ -43,7 +43,9 @@ namespace WebTier.Controllers
                 return BadRequest("A name is required");
             }
 
-            await _repository.UpdateAsync(id, id, builder => builder.Replace(language => language.Name, name));
+            await _repository
+                .UpdateAsync(id: id,
+                    builder: builder => builder.Replace(language => language.Name, name));
 
             return Ok();
         }
--- a/Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
@@ -35,6 +35,19 @@ namespace WebTier.Controllers
         public ValueTask<Language> PutLanguage([FromBody] Language language) =>
             _repository.UpdateAsync(language);
 
+        [HttpPut("{id}")]
+        public async ValueTask<IActionResult> UpdateName([FromRoute] string id, [FromQuery] string name)
+        {
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                return BadRequest("A name is required");
+            }
+
+            await _repository.UpdateAsync(id, id, builder => builder.Replace(language => language.Name, name));
+
+            return Ok();
+        }
+
         [HttpDelete("{id}", Name = nameof(DeleteLanguage))]
         public ValueTask DeleteLanguage(string id) =>
             _repository.DeleteAsync(id);
--- a/Microsoft.Azure.CosmosRepository/src/Builders/IPatchOperationBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/IPatchOperationBuilder.cs
@@ -9,22 +9,24 @@ using Microsoft.Azure.Cosmos;
 namespace Microsoft.Azure.CosmosRepository.Builders
 {
     /// <summary>
-    ///
+    /// Allows a collection of <see cref="PatchOperation"/>'s to built./>
     /// </summary>
     public interface IPatchOperationBuilder<TItem> where TItem : IItem
     {
         /// <summary>
-        ///
+        /// The currently built <see cref="PatchOperation"/>'s
         /// </summary>
         IReadOnlyList<PatchOperation> PatchOperations { get; }
 
         /// <summary>
-        ///
+        /// Allows a property of an <see cref="IItem"/> to be replaced with the value provided
         /// </summary>
-        /// <param name="expression"></param>
-        /// <param name="value"></param>
-        /// <typeparam name="TValue"></typeparam>
-        /// <returns></returns>
+        /// <param name="expression">The expression to define which property to operate on.</param>
+        /// <param name="value">The value to replace the property defined with.</param>
+        /// <typeparam name="TValue">The type of the property that is been replaced.</typeparam>
+        /// <returns>The same instance of <see cref="IPatchOperationBuilder{TItem}"/></returns>
+        /// <remarks>This currently only supports operations on properties on the root level of a JSON document,
+        /// replacing properties on a nested object for example are currently not supported.</remarks>
         IPatchOperationBuilder<TItem> Replace<TValue>(Expression<Func<TItem, TValue>> expression, TValue value);
     }
 }
--- a/Microsoft.Azure.CosmosRepository/src/Builders/IPatchOperationBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/IPatchOperationBuilder.cs
@@ -1,0 +1,30 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.Azure.Cosmos;
+
+namespace Microsoft.Azure.CosmosRepository.Builders
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public interface IPatchOperationBuilder<TItem> where TItem : IItem
+    {
+        /// <summary>
+        ///
+        /// </summary>
+        IReadOnlyList<PatchOperation> PatchOperations { get; }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="expression"></param>
+        /// <param name="value"></param>
+        /// <typeparam name="TValue"></typeparam>
+        /// <returns></returns>
+        IPatchOperationBuilder<TItem> Replace<TValue>(Expression<Func<TItem, TValue>> expression, TValue value);
+    }
+}
--- a/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
+++ b/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.Cosmos.Linq;
+using Microsoft.Azure.CosmosRepository.Builders;
 using Microsoft.Azure.CosmosRepository.Extensions;
 using Microsoft.Azure.CosmosRepository.Options;
 using Microsoft.Azure.CosmosRepository.Processors;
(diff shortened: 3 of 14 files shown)


mumby0168:
Hi, @IEvangelist would not mind your opinion on this still very much WIP at the minute and would currently only support types declared on the root of an object, will look into nested objects and collections soon. Think this will be really cool tho if we can get it right.

The other thing to note is the Cosmos SDK version that currently supports this is in preview so we'd have to hold off releasing this. Also we have jumped to minor versions with the SDK I need to check there is nothing to worry about in terms of our library and the new cosmos SDK updates.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -182,6 +183,19 @@ await container.UpsertItemAsync<TItem>(value, new PartitionKey(value.PartitionKe
             return updateTasks.Select(x => x.Result);
         }
 
+        public async ValueTask UpdateAsync(string id, string partitionKeyValue,
+            Action<IPatchOperationBuilder<TItem>> builder,
+            CancellationToken cancellationToken = default)
```
IEvangelist, line 188:
```suggestion
        public async ValueTask UpdateAsync(
            string id,
            string partitionKeyValue,
            Action<IPatchOperationBuilder<TItem>> builder,
            CancellationToken cancellationToken = default)

```

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -180,7 +196,8 @@ public interface IRepository<TItem> where TItem : IItem
         /// <param name="partitionKeyValue">The partition key value if different than the <see cref="IItem.Id"/>.</param>
         /// <param name="cancellationToken">The cancellation token to use when making asynchronous operations.</param>
         /// <returns>A <see cref="ValueTask"/> representing the asynchronous exists operation.</returns>
-        ValueTask<bool> ExistsAsync(string id, string partitionKeyValue = null, CancellationToken cancellationToken = default);
+        ValueTask<bool> ExistsAsync(string id, string partitionKeyValue = null,
```
IEvangelist, line 199: This is just a nit, but if we wrap some - I prefer wrapping all parameter lists.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -137,6 +138,21 @@ public interface IRepository<TItem> where TItem : IItem
...
+        ///
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="partitionKeyValue"></param>
+        /// <param name="builder"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        ValueTask UpdateAsync(
+            string id,
+            string partitionKeyValue,
```
IEvangelist, line 152:
```suggestion
            string partitionKeyValue = null,
```

I'd imagine that for parity with the other APIs we'd want this to be `null`, and assigned auto-magically like we do elsewhere?

Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
```diff
@@ -35,8 +35,21 @@ public class LanguageController : ControllerBase
...
 
+        [HttpPut("{id}")]
+        public async ValueTask<IActionResult> UpdateName([FromRoute] string id, [FromQuery] string name)
+        {
+            if (string.IsNullOrWhiteSpace(name))
+            {
+                return BadRequest("A name is required");
+            }
+
+            await _repository.UpdateAsync(id, id, builder => builder.Replace(language => language.Name, name));
```
IEvangelist, line 46:
```suggestion
            await _repository.UpdateAsync(
                id, builder: b => b.Replace(language => language.Name, name));
```

Once the `UpdateAsync` API is updated to accept a `null` for the `partionKeyValue` then you could do this, thinking about it a bit more - we might want to change the order to `id, builder, partionKeyValue, cancallationToken`

Microsoft.Azure.CosmosRepository/src/Internals/InternalPatchOperation.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Azure.Cosmos;
+
+namespace Microsoft.Azure.CosmosRepository.Internals
+{
+    internal class InternalPatchOperation
+    {
+        public PatchOperationType Type { get; }
+        public PropertyInfo PropertyInfo { get; }
+
+        public object NewValue { get; }
```
IEvangelist, line 14: I'd prefer if we could somehow still flow the actual generic type `T`, instead of `object` here. I see that their patch operation class has `T`.

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Patch/PatchOperation.cs

IEvangelist (commented):
I love how this is shaping up so far! Looks great.

Microsoft.Azure.CosmosRepository/src/Internals/InternalPatchOperation.cs
mumby0168: This object is purely for the implementation of the in-memory implementation the `InternalPatchOperation` class is not used for actual implementation.

Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
mumby0168: Yes, we will have to re-jig the parameters.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Yes will update this.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Ah yes I did not spot this will update while I am here.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: will do this once I have re-jigged the order of the parameters.

Microsoft.Azure.CosmosRepository.Samples/WebTier/Controllers/LanguageController.cs
mumby0168: Have amended this.

mumby0168:
Need to make sure this is not merged until the SDK supports this fully

IEvangelist:
> Need to make sure this is not merged until the SDK supports this fully

@mumby0168  Should we put it in a separate branch then to avoid potentially shipping something from `main` that isn't ready yet?

Microsoft.Azure.CosmosRepository/src/Extensions/ExpressionExtensions.cs
```diff
@@ -41,5 +42,23 @@ internal static class ExpressionExtensions
...
+        {
+            Type type = typeof(TSource);
+
+            MemberExpression member = propertyLambda.Body as MemberExpression;
+            if (member == null)
+                throw new ArgumentException($"Expression '{propertyLambda}' refers to a method, not a property.");
+
+            PropertyInfo propInfo = member.Member as PropertyInfo;
+            if (propInfo == null)
+                throw new ArgumentException($"Expression '{propertyLambda}' refers to a field, not a property.");
```
IEvangelist, line 56:
```suggestion
            if (propertyLambda.Body is not MemberExpression member)
                throw new ArgumentException($"Expression '{propertyLambda}' refers to a method, not a property.");

            if (member.Member is not PropertyInfo propInfo)
                throw new ArgumentException($"Expression '{propertyLambda}' refers to a field, not a property.");
```

I believe we could use the `is not` operator instead of `as` and `null` check. But you'd have to verify.

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -118,6 +121,39 @@ public async ValueTask<IEnumerable<TItem>> UpdateAsync(IEnumerable<TItem> values
...
+            await Task.CompletedTask;
+
+            partitionKeyValue ??= id;
+
+            IItem item = Items.Values.FirstOrDefault(x => x.Id == id && x.PartitionKey == partitionKeyValue);
+
+            if (item is null)
+            {
+                NotFound();
+            }
```
IEvangelist, line 139:
```suggestion
            IItem item = Items.Values.FirstOrDefault(x => x.Id == id && x.PartitionKey == partitionKeyValue);
            if (item is null)
            {
                NotFound();
            }
```
When assigning a variable, it's preferred to remove the newline immediately following it when it's used straight away.

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -118,6 +121,39 @@ public async ValueTask<IEnumerable<TItem>> UpdateAsync(IEnumerable<TItem> values
...
+            PatchOperationBuilder<TItem> patchOperationBuilder = new();
+
+            builder(patchOperationBuilder);
+
+            foreach (InternalPatchOperation internalPatchOperation in patchOperationBuilder._rawPatchOperations)
+            {
+                if (internalPatchOperation.Type is not PatchOperationType.Replace)
+                {
+                    continue;
+                }
```
IEvangelist, line 150:
```suggestion
            foreach (InternalPatchOperation internalPatchOperation in 
                patchOperationBuilder._rawPatchOperations.Where(ipo => ipo.Type is PatchOperationType.Replace))
            {
```

I'd use a `.Where` clause instead, less noisy. And instead of a negation, serves as a projection.

IEvangelist (commented):
A bit more feedback. Is the plan to add the other patch operations to this PR?

mumby0168:
> > Need to make sure this is not merged until the SDK supports this fully
> 
> @mumby0168 Should we put it in a separate branch then to avoid potentially shipping something from `main` that isn't ready yet?

@IEvangelist I was planning to wait until the sdk has been upgraded, we could put it on another branch, but think it is this or next week they plan to release

mumby0168:
> A bit more feedback. Is the plan to add the other patch operations to this PR?

@IEvangelist I was just going to start with replace operations for now, not sure what place some of the operations have in this library, as we are using a strongly typed object as the contract, what did you think? The other's I had seen was allowing addition and removal of JSON properties as a whole.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -182,6 +183,23 @@ await container.UpsertItemAsync<TItem>(value, new PartitionKey(value.PartitionKe
...
         }
 
+        public async ValueTask UpdateAsync(string id,
+            Action<IPatchOperationBuilder<TItem>> builder,
+            string partitionKeyValue = null,
+            CancellationToken cancellationToken = default)
+        {
+            IPatchOperationBuilder<TItem> patchOperationBuilder = new PatchOperationBuilder<TItem>();
+
+            builder(patchOperationBuilder);
```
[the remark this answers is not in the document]
mumby0168: Not as I have designed it, it just builds up a collection of Patch operations.

Microsoft.Azure.CosmosRepository/src/Builders/IPatchOperationBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public interface IPatchOperationBuilder<TItem> where TItem : IItem
+    {
+        /// <summary>
+        ///
+        /// </summary>
+        IReadOnlyList<PatchOperation> PatchOperations { get; }
+
+        /// <summary>
+        ///
```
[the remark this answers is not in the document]
mumby0168: Yes I forgot about these haha.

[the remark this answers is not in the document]
mumby0168: I have added these in now.

PR #135, opened 2021-11-13
Docs Site Setup

mumby0168:


docs/config.toml
```diff
@@ -0,0 +1,12 @@
+baseURL = 'http://example.org/'
```
mumby0168, line 1: @IEvangelist think we need to update this to be the root of the github pages site do you have the full URL?

IEvangelist (approved):

PR #137, opened 2021-11-14
Optimistic Concurrency - etag Support

### LINKED ISSUE #83: Transactions and optimistic concurrency control
Hi I am working on a project that is using your library, the project requires that we handle some form of Transactions or optimistic concurrency control on the data. Essentially I need to read the data and change contents within the document and write it back.

What I want to avoid is when writing back to have a newer version in the database and my data ends up overwriting the current document. 

Microsoft has already something that handles such situations (https://docs.microsoft.com/en-us/azure/cosmos-db/database-transactions-optimistic-concurrency#optimistic-concurrency-control) is this implemented within your library?

mumby0168:
Closes #83


mumby0168:
Hi @IEvangelist I have been working on adding support for `etags` to be used with an `IItem` I thought about just adding an extra field to `IItem` called `Etag` which would do the trick, however i'd prefer to not extend this class anymore and maybe look at doing something a little smarter, the plan is to do something like this below

```csharp
options.Builder..ConfigureOptions<User>(optionsBuilder => optionsBuilder.WithETag("_privateEtagField"));
//or
options.Builder..ConfigureOptions<User>(optionsBuilder => optionsBuilder.WithETag(x => x.MyCustomEtagProperty));
```

which is what this PR has scoped out a little, the only problem been is that when you perform a `ReadItemAsync<T>()` you get the `etag` property is available on the `FeedResponse<T> response` via the `response.ETag` property which is not a problem, we could then just use reflection to the set the `etag` property on whatever objects the client of the library has defined.

The problem comes when we are reading a collection, we would have to first read the object into a `dynamic` object when calling `ReadNextAsync()` then we could access the `etag` field using `dynamicItem.etag` and then set this with reflection then I have been doing some investigation we could then deserialize the object again into the strong type and add it to the list as we do now. The problem I have is that I am not sure what extra overhead or performance hit this would incur if we decided to do this? The little sample I did to prove this works is below

```csharp
var john = new Person("123", "John");

var settings = new JsonSerializerOptions
{
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
    WriteIndented = true,
    PropertyNameCaseInsensitive = true
};

var json = JsonSerializer.Serialize(john, settings);

Console.WriteLine(json);

var dynamicPerson = JsonSerializer.Deserialize<dynamic>(json, settings);

Person newPerson = JsonSerializer.Deserialize<Person>(dynamicPerson, settings);

Console.WriteLine(newPerson);
record Person(string Id, string Name);

// Output
// {
//   "id": "123",
//   "name": "John"
// }
// Person { Id = 123, Name = John }

```

I was wondering your thought's on the current ideas before I spend any more time down this route, i'd like the idea of being able to set and get on private fields as we have been trying to use this library while implementing DD and domain objects and it has been tricky have some extra properties on `IItem` that we have to pollute our domain objects with. I thought we could do something similar for the setting the last updated time and maybe even the partition key value if we get this right here then we could end up with `IItem` just having the `Id` property, again interested to hear your opinion 😃

Microsoft.Azure.CosmosRepository/src/Builders/ItemOptionsBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.Configuration.Internal;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Microsoft.Azure.CosmosRepository.Builders
+{
+    /// <summary>
+    /// Allows metadata such etags and last updated times to be added to an <see cref="IItem"/>
```
IEvangelist, line 12:
```suggestion
    /// Allows metadata such as eTags and last updated times to be added to an <see cref="IItem"/>
```

Microsoft.Azure.CosmosRepository/src/Builders/ItemOptionsBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// </summary>
+    public class ItemOptionsBuilder
+    {
+        internal ItemOptionsBuilder(Type type)
+        {
+            Type = type;
+            ETagPropertyInformation = null;
+        }
+        internal Type Type { get; private set; }
```
IEvangelist, line 21:
```suggestion
        
        internal Type Type { get; private set; }
```

Microsoft.Azure.CosmosRepository/src/Builders/IItemBuilder.cs
```diff
@@ -11,18 +11,26 @@ namespace Microsoft.Azure.CosmosRepository.Builders
...
+        /// <returns>Instance of <see cref="IItemBuilder"/></returns>
+        IItemBuilder ConfigureContainer<TItem>(Action<ContainerOptionsBuilder> containerOptions) where TItem : IItem;
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="metadataBuilder"></param>
+        /// <typeparam name="TItem"></typeparam>
+        /// <returns></returns>
+        IItemBuilder ConfigureOptions<TItem>(Action<ItemOptionsBuilder> metadataBuilder) where TItem : IItem;
```
IEvangelist, line 34: I feel like the name `ConfigureOptions` is a bit too vague. Perhaps something to more clearly indicate that it lets you configure metadata, such as eTag, etc. Does `ConfigureItemPropertyMetadata` make more sense? I'm not really sure I like that name either. LOL

IEvangelist (commented):
This looks good so far, but I'm not sure where the configured etag options will be consumed. Is that still to come?

mumby0168:
> This looks good so far, but I'm not sure where the configured etag options will be consumed. Is that still to come?

Yes still yet to come, just wanted to make sure you were happy with the intention for this

Microsoft.Azure.CosmosRepository/src/Builders/IItemBuilder.cs
mumby0168: Yes I started with with `ConfigureMetadata(...)` but then thought against it haha, what do you think?

IEvangelist:
> > This looks good so far, but I'm not sure where the configured etag options will be consumed. Is that still to come?
> 
> Yes still yet to come, just wanted to make sure you were happy with the intention for this

Ok, I agree that the double serialization would be less than ideal. Would it be better to create a serializer interceptor of sorts, that acts as a pass-thru for all standard and non-etag configure items, but when it detects and object type that has specified this special behavior - only then does it map over the value of the etag?

I'm honestly not sure this would work. But if it would you'd probably have much more involved, you'd have to intercept any existing serialization settings, evaluate each type being serialized, check if anything is configured, etc. Does that make sense? This is the only idea that I have thus far that should work in a single pass. Thoughts?

mumby0168:
> > > This looks good so far, but I'm not sure where the configured etag options will be consumed. Is that still to come?
> > 
> > 
> > Yes still yet to come, just wanted to make sure you were happy with the intention for this
> 
> Ok, I agree that the double serialization would be less than ideal. Would it be better to create a serializer interceptor of sorts, that acts as a pass-thru for all standard and non-etag configure items, but when it detects and object type that has specified this special behavior - only then does it map over the value of the etag?
> 
> I'm honestly not sure this would work. But if it would you'd probably have much more involved, you'd have to intercept any existing serialization settings, evaluate each type being serialized, check if anything is configured, etc. Does that make sense? This is the only idea that I have thus far that should work in a single pass. Thoughts?

Okay I like that idea, not used interceptors at all before any resources you'd recommend?

mumby0168:
Another thought on this @IEvangelist is making using of a `JObject` this may be simpler than the interceptors what do you think?

IEvangelist:
> Another thought on this @IEvangelist is making using of a `JObject` this may be simpler than the interceptors what do you think?

Can a `JObject` be converted to a strongly-typed object without another deserialization invocation? Here is what I was thinking of:

https://www.newtonsoft.com/json/help/html/CustomJsonConverterGeneric.htm

Ideally, we could create something that defaults to the Newtonsoft JSON but allows us to change serializers easily. Perhaps an abstraction?

IEvangelist:
> > > Another thought on this @IEvangelist is making using of a `JObject` this may be simpler than the interceptors what do you think?
> > 
> > 
> > Can a `JObject` be converted to a strongly-typed object without another deserialization invocation? Here is what I was thinking of:
> > https://www.newtonsoft.com/json/help/html/CustomJsonConverterGeneric.htm
> > Ideally, we could create something that defaults to the Newtonsoft JSON but allows us to change serializers easily. Perhaps an abstraction?
> 
> Not sure I am following this thought to well, I don't think we have much control of the serializer used by the cosmos SDK in this instance, what would the `T` be on the `JsonConverter<T>` ?

I assumed that since the Cosmos library was using Newtonsoft, that we could define a `JsonConverter<T>` on our base `Item` class. The we could configure the client to use it, again - I'm not sure if that is possible. I did just find this: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.cosmosserializer?view=azure-dotnet We have access to the raw stream. We could hypothetically use `System.Text.Json` with these raw stream values, and add any appropriate logic for managing the etag (if configured) for a given `T`.

mumby0168:
> > > > Another thought on this @IEvangelist is making using of a `JObject` this may be simpler than the interceptors what do you think?
> > > 
> > > 
> > > Can a `JObject` be converted to a strongly-typed object without another deserialization invocation? Here is what I was thinking of:
> > > https://www.newtonsoft.com/json/help/html/CustomJsonConverterGeneric.htm
> > > Ideally, we could create something that defaults to the Newtonsoft JSON but allows us to change serializers easily. Perhaps an abstraction?
> > 
> > 
> > Not sure I am following this thought to well, I don't think we have much control of the serializer used by the cosmos SDK in this instance, what would the `T` be on the `JsonConverter<T>` ?
> 
> I assumed that since the Cosmos library was using Newtonsoft, that we could define a `JsonConverter<T>` on our base `Item` class. The we could configure the client to use it, again - I'm not sure if that is possible. I did just find this: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.cosmosserializer?view=azure-dotnet We have access to the raw stream. We could hypothetically use `System.Text.Json` with these raw stream values, and add any appropriate logic for managing the etag (if configured) for a given `T`.

Ahhh I like this it would work really nicely as well, i'll see if I get some time to experiment this week

mumby0168:
@IEvangelist might try doing something similar to this

https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Serializer/CosmosJsonDotNetSerializer.cs

mumby0168:
After the discussion with the EF core team, I think this may be overkill & I think I favour just adding an `ETag` property to the `Item` base class.

PR #142, opened 2021-11-21
Paging (Scroll) Continuation Token Tupport

mumby0168:
Related to:

- #57
- #82


Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -198,6 +199,12 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
                 item => item.Type == typeof(TItem).Name, Expression.AndAlso).Compile());
         }
 
+        public ValueTask<IPage<TItem>> ScrollAsync(int lastPage = 0, Expression<Func<TItem, bool>> predicate = null, int pageSize = 25, string continuationToken = null,
+            CancellationToken cancellationToken = default)
```
IEvangelist, line 203:
```suggestion
        public ValueTask<IPage<TItem>> ScrollAsync(
            int page = 0,
            Expression<Func<TItem, bool>> predicate = null,
            int pageSize = 25,
            string continuationToken = null,
            CancellationToken cancellationToken = default)
```

Format nit, and parameter name suggestion.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +279,39 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
...
+            };
+
+            IQueryable<TItem> query = container.GetItemLinqQueryable<TItem>(requestOptions: options, continuationToken: continuationToken)
+                .Where(_repositoryExpressionProvider.Build(predicate));
+
+            Response<int> countResponse = await query.CountAsync(cancellationToken);
+
+            List<TItem> results = new();
+
+            FeedIterator<TItem> iterator = query.ToFeedIterator();
```
IEvangelist, line 302: Are you missing a `using` here, I'm not sure - you'd have to verify (also check if it is an imple?

```suggestion
            using FeedIterator<TItem> iterator = query.ToFeedIterator();
```

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +279,39 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
             return count > 0;
         }
 
+        public async ValueTask<IPage<TItem>> ScrollAsync(int lastPage = 0,
+            Expression<Func<TItem, bool>> predicate = null,
```
IEvangelist, line 283:
```suggestion
        public async ValueTask<IPage<TItem>> ScrollAsync(
            int lastPage = 0,
            Expression<Func<TItem, bool>> predicate = null,
```

Format nit

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +279,39 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
             return count > 0;
         }
 
+        public async ValueTask<IPage<TItem>> ScrollAsync(int lastPage = 0,
```
IEvangelist, line 282: I'm curious how this makes @roji feel? What do you think about this API and implementation?

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -219,5 +220,20 @@ public interface IRepository<TItem> where TItem : IItem
         /// <returns>A <see cref="ValueTask"/> representing the asynchronous exists operation.</returns>
         ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        ///
```
IEvangelist, line 225: Missing comments, TODO: 😬

IEvangelist (commented):
I love the way this is shaping up so far!

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Will add these now.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: Yes very keen to hear some feedback 😃

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Added these

Microsoft.Azure.CosmosRepository/src/Paging/Page.cs
```diff
@@ -0,0 +1,51 @@
...
+        public Page(int total, int number, int size, IEnumerable<T> items, double charge, string? continuation = null)
+        {
+            Total = total;
+            Number = number;
+            Size = size;
+            Items = items;
+            Charge = charge;
+            Continuation = continuation;
+        }
+
```
IEvangelist, line 31:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/Paging/Page.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.CosmosRepository.Paging
+{
+
```
IEvangelist, line 9:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/Paging/IPage.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// The size of the current page.
+        /// </summary>
+        int Size { get; }
+
+        /// <summary>
+        /// The items that are in the current page.
+        /// </summary>
+        IEnumerable<T> Items { get; }
```
IEvangelist, line 33:
```suggestion
        IReadOnlyList<T> Items { get; }
```

Since we are materializing a list, perhaps we should return an `IReadOnlyList`?

Microsoft.Azure.CosmosRepository/src/Paging/Page.cs
```diff
@@ -0,0 +1,51 @@
...
+        public int Total { get; }
+
+        /// <inheritdoc />
+        public int Number { get; }
+
+        /// <inheritdoc />
+        public int Size { get; }
+
+        /// <inheritdoc />
+        public IEnumerable<T> Items { get; }
```
IEvangelist, line 43:
```suggestion
        public IReadOnlyList<T> Items { get; }
```

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +279,40 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
...
+
+            FeedResponse<TItem> next = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
+
+            foreach (TItem result in next)
+            {
+                results.Add(result);
+            }
+
+            lastPage++;
+            return new Page<TItem>(countResponse.Resource, lastPage, pageSize, results, next.RequestCharge + countResponse.RequestCharge, next.ContinuationToken);
```
IEvangelist, line 313:
```suggestion
            lastPage++;
            return new Page<TItem>(
                countResponse.Resource,
                lastPage,
                pageSize,
                results.AsReadOnly(),
                next.RequestCharge + countResponse.RequestCharge,
                next.ContinuationToken);
```

Microsoft.Azure.CosmosRepository/src/Paging/IPage.cs
mumby0168: Yes I think that makes sense

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +279,46 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
             return count > 0;
         }
 
+        public async ValueTask<IPage<TItem>> ScrollAsync(
+            int lastPage = 0,
```
roji, line 283: This parameter seems a bit odd - it seems to only be incremented and integrated in the returned Page (no actual effect on the query etc). I'm not sure I'd do this rather than leaving it up to the API user to track which page their own (at least with the current API shape) - but see my general comment about an API which streams pages.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -219,5 +220,22 @@ public interface IRepository<TItem> where TItem : IItem
...
         ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Offers a load more paging implementation for infinite scroll scenarios.
+        /// Allows for efficient paging making use of cosmos DBs continuation tokens, making this implementation cost effective.
+        /// </summary>
+        /// <param name="lastPage">The last page previous query, defaults to 0 for the first page</param>
+        /// <param name="predicate">A filter criteria for the paging operation, if null it will get all <see cref="IItem"/>s</param>
+        /// <param name="pageSize">The size of the page to return from cosmos db.</param>
```
roji, line 230: Don't forget that MaxItemCount is only a maximum - Cosmos may send back less items. I think there's two possible options here:

1. Expose this to the user by renaming `pageSize` to `maxPageSize` and clearly documenting it. It then becomes the user's responsibility to call ScrollAsync repeatedly until they get enough items to have a full page.
2. Loop internally yourself until you get the requested pageSize.

I suspect that for most actual real-world usages, (2) would be more useful - but I'm not sure how high-level/low-level an API you're looking to create here.

roji (commented):
Sorry it took me such a long time to look at this.

It looks good - it's nice to see continuation tokens being used like this.

I have another possible API shape in mind... The current API has the user repeatedly calling ScrollAsync to get more results; it makes it up to them to remember the continuation token between invocations, and also to manage the current/last page number (which IMHO is quite weird currently, see comment below).

Instead, you could expose an API which returns `IAsyncEnumerable<IPage>`; the enumerator would contain the continuation internally (the user would never have to see/deal with it), and could even expose a `LastPage` or `CurrentPage` property for the user. This would assume that the page size doesn't change between pages, which seems right to me.

mumby0168:
> Sorry it took me such a long time to look at this.
> 
> It looks good - it's nice to see continuation tokens being used like this.
> 
> I have another possible API shape in mind... The current API has the user repeatedly calling ScrollAsync to get more results; it makes it up to them to remember the continuation token between invocations, and also to manage the current/last page number (which IMHO is quite weird currently, see comment below).
> 
> Instead, you could expose an API which returns `IAsyncEnumerable<IPage>`; the enumerator would contain the continuation internally (the user would never have to see/deal with it), and could even expose a `LastPage` or `CurrentPage` property for the user. This would assume that the page size doesn't change between pages, which seems right to me.

Hi @roji thanks for your review 😃 

So the reason I chose to return the CT as part of the `IPage<T>` was thinking more of an API usage so a user has a web app they want to scroll through a set of pages they won't be in the code using the method so they'll need to pass the CT back to us. I agree it's a little odd I thought the same, the only over alternative was to cache the CT token which adds a lot of complexity. It might suit to add an implementation that has a different set of arguments for the scenarios where a user may scroll in code multiple pages using `IAsyncEnumerble<T>`, what do you think?

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: Yes, this did seem a little odd I suppose we could remove this parameter and like you say let the caller keep track of the current page. As this does not really serve any purpose here.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: I had assumed that a user would expect fewer items if there is not enough data in cosmos to satisfy the page size, was this what you meant, or did you mean we may have to do multiple calls to `ReadNextAsync()` ?

roji: So the way I'm reading [the docs](https://docs.microsoft.com/en-us/azure/cosmos-db/sql/sql-query-pagination), Cosmos may return less than MaxItemCount even if there are enough items in the resultset:

> The number of items returned per query execution will always be less than or equal to MaxItemCount. However, it is possible that other criteria might have limited the number of results the query could return. If you execute the same query multiple times, the number of pages might not be constant. For example, if a query is throttled there may be fewer available results per page, which means the query will have additional pages. In some cases, it is also possible that your query may return an empty page of results.

Now, if the main usage for this API is for a paging UI (I suspect that's the case), then the user wants to always fill out e.g. 25 items in each page (except the last). The way the code is currently structured, I think they'd need to call ScrollAsync repeatedly until they have the 25 items (this also complicates things for the page count, BTW). My (2) proposal above would be to do that loop inside ScrollAsync, guaranteeing that it always returns 25 items, unless there are no more results at all.

Note also that Cosmos may return 0 items as well, although there are items available. Keep this in mind with regards to detecting the end of the query resultset.

roji:
> So the reason I chose to return the CT as part of the IPage<T> was thinking more of an API usage so a user has a web app they want to scroll through a set of pages they won't be in the code using the method so they'll need to pass the CT back to us.

I think you're referring to a disconnected web scenario, where the continuation token needs to be passed all the way back to the client, and keeping state such as an IAsyncEnumerator may not be appropriate - that's indeed true. Maybe there can be two APIs here: the first low-level one is more or less what you already have (for disconnected scenarios), and a second, higher-level one could be built on top of the first, returning an enumerable...

mumby0168:
> > So the reason I chose to return the CT as part of the IPage was thinking more of an API usage so a user has a web app they want to scroll through a set of pages they won't be in the code using the method so they'll need to pass the CT back to us.
> 
> I think you're referring to a disconnected web scenario, where the continuation token needs to be passed all the way back to the client, and keeping state such as an IAsyncEnumerator may not be appropriate - that's indeed true. Maybe there can be two APIs here: the first low-level one is more or less what you already have (for disconnected scenarios), and a second, higher-level one could be built on top of the first, returning an enumerable...
> 
> The latter would be useful for all kinds of stateful users, e.g. Winforms, Blazor...

@roji  Yes I really like that idea, I will add an issue for and like you say we can look to build this on top of the API we already have, we plan to implement the skip, take an approach for paging as well, with the disclaimer we spoke of around performance.

Thanks again for taking the time to review, much appreciated.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Ah okay this makes sense now, yes look's as if we will need to be more intelligent here, I will update this shortly and push the changes.

mumby0168: Hi @roji finally got around to updating this, think this should do the trick now 😃

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: Removed this parameters as it did not provide much benefit 👍

.github/workflows/codeql-analysis.yml
```diff
@@ -58,6 +61,13 @@ jobs:
         # Prefix the list here with "+" to use these queries and those in the config file.
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: | 
+                  3.1.402
+                  6.0.x
```
IEvangelist, line 69:
```suggestion
        dotnet-version: | 
          3.1.402
          6.0.x
```

.github/workflows/pre-release-nuget.yml
```diff
@@ -25,8 +25,9 @@ jobs:
     - uses: actions/checkout@v2
     - name: Setup .NET  
       uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 3.1.402
+      dotnet-version: | 
+          3.1.402
+          6.0.x
```
IEvangelist, line 30:
```suggestion
      with:
        dotnet-version: | 
          3.1.402
          6.0.x
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Paging.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <LangVersion>latest</LangVersion>
+        <UserSecretsId>9c9f04ba-0288-4b94-a15e-c669da85dfad</UserSecretsId>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Bogus" Version="33.0.2" />
+      <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.2" />
```
IEvangelist, line 12: If you're TFM is `net6.0`, your `PackageReference` could probably target that too.

Microsoft.Azure.CosmosRepository.Samples/Paging/Person.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosRepository;
+using Microsoft.Azure.CosmosRepository.Attributes;
+
+[Container("people")]
+[PartitionKeyPath("/id")]
+
```
IEvangelist, line 9:
```suggestion
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+        options.DatabaseId = "paging-db";
+        options.ContainerPerItemType = true;
+    })
+    .AddSingleton<IConfiguration>(configuration.Build())
+    .BuildServiceProvider();
+
+IRepository<Person> repository = provider.GetRequiredService<IRepository<Person>>();
+
```
IEvangelist, line 23:
```suggestion
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+await Seed();
+await BasicScrolling();
```
IEvangelist, line 26:
```suggestion
await SeedAsync();
await BasicScrollingAsync();
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+async Task BasicScrolling()
```
IEvangelist, line 28:
```suggestion
async Task BasicScrollingAsync()
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    totalCharge += page.Charge;
+
+    Console.WriteLine($"Last 50 results cost {page.Charge}");
+
+    Console.WriteLine($"Total Charge {totalCharge} RU's");
+}
+
+async Task Seed()
```
IEvangelist, line 67:
```suggestion
async Task SeedAsync()
```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    foreach (Person person in page.Items)
+    {
+        Console.WriteLine(person);
+    }
+
+    totalCharge += page.Charge;
+
+    Console.WriteLine($"Last 50 results cost {page.Charge}");
+
```
IEvangelist, line 63:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -74,7 +74,7 @@ await container.ReadItemAsync<TItem>(id, partitionKey, cancellationToken: cancel
 
             TryLogDebugDetails(_logger, () => $"Read: {JsonConvert.SerializeObject(item)}");
 
-            return item is { Type: { Length: 0 } } || item.Type == typeof(TItem).Name ? item : default;
+            return item is {Type: {Length: 0}} || item.Type == typeof(TItem).Name ? item : default;
```
IEvangelist, line 77:
```suggestion
            return item is { Type: { Length: 0 } } || item.Type == typeof(TItem).Name ? item : default;

```

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +284,60 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
             return count > 0;
         }
 
+        public async ValueTask<IPage<TItem>> ScrollAsync(
```
IEvangelist, line 287: Thinking a bit more about the exposed API, should we call this "PageAsync" instead? The word "Scroll" might seem to imply the intended use case such as a UI. What do you think?

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -279,6 +284,60 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
...
+
+            List<TItem> results = new();
+
+            int readItemsCount = 0;
+
+            using FeedIterator<TItem> iterator = query.ToFeedIterator();
+
+            double charge = countResponse.RequestCharge;
+
+            while (readItemsCount < pageSize && iterator.HasMoreResults)
```
IEvangelist, line 314:
```suggestion
            Response<int> countResponse = await query.CountAsync(cancellationToken);
            using FeedIterator<TItem> iterator = query.ToFeedIterator();

            List<TItem> results = new();
            int readItemsCount = 0;
            double charge = countResponse.RequestCharge;

            while (readItemsCount < pageSize && iterator.HasMoreResults)
```

I think that some of the extra whitespace and new lines made it harder to read. I grouped a few things together and slightly re-orged it a bit.

IEvangelist (dismissed):
Looks great, well done kind sir. I have a few nits, and a general question. Otherwise, let's ship it.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: Looks good to me

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: I initially went for scroll, due to us aiming to implement paging using offset just wondering if that could be achieved by an overload to `PageAsync` I am easy either way, happy for you to make the call on this one :)

Microsoft.Azure.CosmosRepository.Samples/Paging/Paging.csproj
mumby0168: Happy to bump it :)

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
mumby0168: Actually, I like this we can do the skip take approach with an API looking like this should fit nicely

```csharp
ValueTask<IPage<TItem>> PageAsync(
            Expression<Func<TItem, bool>> predicate = null,
            int pageNumber = 1,
            int pageSize = 25,
```

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -43,7 +43,7 @@ internal class DefaultRepository<TItem> : IRepository<TItem> where TItem : IItem
             ICosmosQueryableProcessor cosmosQueryableProcessor,
             IRepositoryExpressionProvider repositoryExpressionProvider) =>
             (_optionsMonitor, _containerProvider, _logger, _cosmosQueryableProcessor, _repositoryExpressionProvider) =
-                (optionsMonitor, containerProvider, logger, cosmosQueryableProcessor, repositoryExpressionProvider);
+            (optionsMonitor, containerProvider, logger, cosmosQueryableProcessor, repositoryExpressionProvider);
```
IEvangelist, line 46:
```suggestion
                (optionsMonitor, containerProvider, logger, cosmosQueryableProcessor, repositoryExpressionProvider);

```

Microsoft.Azure.CosmosRepository.Samples/Paging/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        options.DatabaseId = "paging-db";
+        options.ContainerPerItemType = true;
+    })
+    .AddSingleton<IConfiguration>(configuration.Build())
+    .BuildServiceProvider();
+
+IRepository<Person> repository = provider.GetRequiredService<IRepository<Person>>();
+
+await SeedAsync();
+await BasicScrolling();
```
IEvangelist, line 25:
```suggestion
await BasicScrollingAsync();
```

IEvangelist (approved):
A few missed nits, otherwise let's :shipit:

mumby0168:
Think this is all good to go if you are happy @IEvangelist

IEvangelist (approved):

mumby0168:
> Looks good to me, we'll :shipit: - thank you 🙏

Great news, the long-awaited feature 🥳

Issue #78, opened 2021-08-04
Last Updated Property on Item


mumby0168:
**Is your feature request related to a problem? Please describe.**
`n/a`

**Describe the solution you'd like**
A property adding to the default `Item` type that includes the `_ts` (last-updated) timestamp provided by cosmos. This can be converted from unix time to a UTC value represented as a `DateTime` with the property name of `LastUpdatedUtc`.

**Describe alternatives you've considered**
Manually adding this property to items that already derive from `Item`

**Additional context**
`n/a`

mumby0168:
Have dupe for this #92

PR #147, opened 2021-12-15
Optimistic concurrency control

robertbennett1998:
Closes: #137

### DIFF
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
@@ -14,7 +14,6 @@
 
     <ItemGroup>
       <PackageReference Include="Bogus" Version="33.0.2" />
-      <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="5.0.0" />
     </ItemGroup>
 
 </Project>
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
@@ -1,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <UserSecretsId>fd4da918-55af-4010-9ba0-e75f0e0dc917</UserSecretsId>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\Microsoft.Azure.CosmosRepository\src\Microsoft.Azure.CosmosRepository.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Bogus" Version="33.0.2" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="5.0.0" />
+    </ItemGroup>
+
+</Project>
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
@@ -1,0 +1,83 @@
+﻿using System.Net;
+using Bogus;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Azure.CosmosRepository;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using OptimisticConcurrencyControl;
+
+ConfigurationBuilder configuration = new();
+
+ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
+    {
+        options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+        options.DatabaseId = "optimistic-concurrency-control";
+        options.OptimizeBandwidth = false;
+    })
+    .AddSingleton<IConfiguration>(configuration.Build())
+    .BuildServiceProvider();
+
+IRepository<BankAccount> repository = provider.GetRequiredService<IRepository<BankAccount>>();
+
+await SeedAsync();
+await Concurrency();
+
+async Task Concurrency()
+{
+    double totalCharge = 0;
+
+    BankAccount createdBankAccount = await repository.CreateAsync(new BankAccount()
+    {
+        Name = "Current Account",
+        Balance = 500.0
+    });
+
+    Console.WriteLine($"Created a bank account: {createdBankAccount}.");
+    Console.WriteLine("Attempting to update the same instance of the created bank account three times.");
+
+    Console.WriteLine($"1) Attempting to update balance to 750.0.");
+    BankAccount bankAccountUpdate = await repository.GetAsync(createdBankAccount.Id);
+    bankAccountUpdate.Balance = 750.0;
+    BankAccount updatedBankAccount = await repository.UpdateAsync(bankAccountUpdate);
+
+    Console.WriteLine($"1) Bank account updated {updatedBankAccount}.");
+
+    try
+    {
+        bankAccountUpdate.Balance = 1000.0;
+        Console.WriteLine($"2) Attempting to update balance to 1000.0 using the original etag. {bankAccountUpdate}");
+        await repository.UpdateAsync(bankAccountUpdate);
+    }
+    catch (CosmosException exception)
+    {
+        if (exception.StatusCode == HttpStatusCode.PreconditionFailed)
+        {
+            Console.WriteLine("2) Failed to update balance to 1000.0 as etags did not match.");
+        }
+    }
+
+    Console.WriteLine($"3) Forcing balance to update update balance to 1000.0 using the original etag. But explicitly telling it to ignore the etag {bankAccountUpdate}");
+    updatedBankAccount = await repository.UpdateAsync(bankAccountUpdate, ignoreEtag: true);
+    Console.WriteLine($"3) Bank account updated {updatedBankAccount}.");
+
+    BankAccount finalAccountInfo = await repository.GetAsync(createdBankAccount.Id);
+    Console.WriteLine($"Final account information {finalAccountInfo}");
+}
+
+async Task SeedAsync()
+{
+    IEnumerable<BankAccount> current = await repository.GetAsync(x => x.Type == nameof(BankAccount));
+
+    if (current.Any())
+    {
+        return;
+    }
+
+    Faker<BankAccount> peopleFaker = new();
+    peopleFaker
+        .RuleFor(p => p.Name, f => f.Name.FullName())
+        .RuleFor(p => p.Balance, f => f.Random.Number(0, 10000));
+
+    List<BankAccount> people = peopleFaker.Generate(100);
+    await repository.CreateAsync(people);
+}
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
@@ -12,7 +12,7 @@ ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =
     {
         options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
         options.DatabaseId = "optimistic-concurrency-control";
-        options.OptimizeBandwidth = false;
+        options.OptimizeBandwidth = false; // Must be false to receive the upto date etags back on update calls
     })
     .AddSingleton<IConfiguration>(configuration.Build())
     .BuildServiceProvider();
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
@@ -59,16 +59,12 @@ async Task ConcurrencyWithOptimizeBandwidthOff()
         await repository.UpdateAsync(currentBankAccount.Id,
             builder => builder.Replace(account => account.Balance, currentBankAccount.Balance - 250), etag: currentBankAccount.Etag);
     }
-    catch (CosmosException exception)
+    catch (CosmosException exception) when (exception.StatusCode == HttpStatusCode.PreconditionFailed)
     {
-        if (exception.StatusCode == HttpStatusCode.PreconditionFailed)
-        {
-            Console.WriteLine("Failed to withdraw 250 from the bank account as the local state was out of date with the remote state.");
-            Console.WriteLine("Syncing local state with remote state.");
-            currentBankAccount = await repository.GetAsync(currentBankAccount.Id);
-            Console.WriteLine($"Updated local state. {currentBankAccount}");
-
-        }
+        Console.WriteLine("Failed to withdraw 250 from the bank account as the local state was out of date with the remote state.");
+        Console.WriteLine("Syncing local state with remote state.");
+        currentBankAccount = await repository.GetAsync(currentBankAccount.Id);
+        Console.WriteLine($"Updated local state. {currentBankAccount}");
     }
 
     try
@@ -124,16 +120,13 @@ async Task ConcurrencyWithOptimizeBandwidthOn()
             builder => builder.Replace(account => account.Balance, currentBankAccount.Balance - 250), etag: currentBankAccount.Etag);
         currentBankAccount = await repository.GetAsync(bankAccountInfo.Id);
     }
-    catch (CosmosException exception)
+    catch (CosmosException exception) when (exception.StatusCode == HttpStatusCode.PreconditionFailed)
     {
-        if (exception.StatusCode == HttpStatusCode.PreconditionFailed)
-        {
-            Console.WriteLine("Failed to withdraw 250 from the bank account as the local state was out of date with the remote state.");
-            Console.WriteLine("Syncing local state with remote state.");
-            currentBankAccount = await repository.GetAsync(currentBankAccount.Id);
-            Console.WriteLine($"Updated local state. {currentBankAccount}");
+        Console.WriteLine("Failed to withdraw 250 from the bank account as the local state was out of date with the remote state.");
+        Console.WriteLine("Syncing local state with remote state.");
+        currentBankAccount = await repository.GetAsync(currentBankAccount.Id);
+        Console.WriteLine($"Updated local state. {currentBankAccount}");
 
-        }
     }
 
     Console.WriteLine("Attempting to withdraw 750 from the bank account.");
@@ -191,12 +184,9 @@ async Task ConcurrencyWithOptimizeBandwidthOnTrapDemo()
         currentBankAccount = await repository.UpdateAsync(currentBankAccount);
         Console.WriteLine($"Updated bank account: {currentBankAccount}.");
     }
-    catch (CosmosException exception)
+    catch (CosmosException exception) when (exception.StatusCode == HttpStatusCode.PreconditionFailed)
     {
-        if (exception.StatusCode == HttpStatusCode.PreconditionFailed)
-        {
-            Console.WriteLine("Failed to update balance as the etags did not match.");
-        }
+        Console.WriteLine("Failed to update balance as the etags did not match.");
     }
 
     Console.WriteLine($"Reattempting using value from the database.");
--- a/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
+++ b/Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
@@ -6,78 +6,198 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using OptimisticConcurrencyControl;
 
-ConfigurationBuilder configuration = new();
 
-ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
-    {
-        options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
-        options.DatabaseId = "optimistic-concurrency-control";
-        options.OptimizeBandwidth = false; // Must be false to receive the upto date etags back on update calls
-    })
-    .AddSingleton<IConfiguration>(configuration.Build())
-    .BuildServiceProvider();
-
-IRepository<BankAccount> repository = provider.GetRequiredService<IRepository<BankAccount>>();
-
-await SeedAsync();
-await Concurrency();
-
-async Task Concurrency()
+IRepository<BankAccount> BuildRepository(bool optimizeBandwidth)
 {
-    double totalCharge = 0;
+    ConfigurationBuilder configuration = new();
 
-    BankAccount createdBankAccount = await repository.CreateAsync(new BankAccount()
+    ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
+        {
+            options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+            options.DatabaseId = "optimistic-concurrency-control";
+            options.OptimizeBandwidth = optimizeBandwidth; // Must be false to receive the upto date etags back on update calls
+        })
+        .AddSingleton<IConfiguration>(configuration.Build())
+        .BuildServiceProvider();
+
+    IRepository<BankAccount> repository = provider.GetRequiredService<IRepository<BankAccount>>();
+
+    return repository;
+}
+
+await ConcurrencyWithOptimizeBandwidthOff();
+await ConcurrencyWithOptimizeBandwidthOn();
+await ConcurrencyWithOptimizeBandwidthOnTrapDemo();
+
+async Task ConcurrencyWithOptimizeBandwidthOff()
+{
+    Console.WriteLine($"Optimized bandwidth OFF.");
+    IRepository<BankAccount> repository = BuildRepository(false);
+
+    BankAccount currentBankAccount = await repository.CreateAsync(new BankAccount()
     {
         Name = "Current Account",
         Balance = 500.0
     });
 
-    Console.WriteLine($"Created a bank account: {createdBankAccount}.");
-    Console.WriteLine("Attempting to update the same instance of the created bank account three times.");
+    Console.WriteLine($"Created a bank account: {currentBankAccount}.");
 
-    Console.WriteLine($"1) Attempting to update balance to 750.0.");
-    BankAccount bankAccountUpdate = await repository.GetAsync(createdBankAccount.Id);
-    bankAccountUpdate.Balance = 750.0;
-    BankAccount updatedBankAccount = await repository.UpdateAsync(bankAccountUpdate);
+    currentBankAccount.Withdraw(500);
+    await repository.UpdateAsync(currentBankAccount);
+    currentBankAccount.Deposit(500);
 
-    Console.WriteLine($"1) Bank account updated {updatedBankAccount}.");
+    Console.WriteLine($"Simulating a withdrawal of 500 that happened from another service. Remote State - {await repository.GetAsync(currentBankAccount.Id)}. Local State - {currentBankAccount}");
 
+    Console.WriteLine("Attempting to withdraw 250 from the bank account.");
     try
     {
-        bankAccountUpdate.Balance = 1000.0;
-        Console.WriteLine($"2) Attempting to update balance to 1000.0 using the original etag. {bankAccountUpdate}");
-        await repository.UpdateAsync(bankAccountUpdate);
+        currentBankAccount.Withdraw(250.0);
+        Console.WriteLine("Attempting to withdraw 250 from the bank account.");
+        currentBankAccount = await repository.UpdateAsync(currentBankAccount);
     }
     catch (CosmosException exception)
     {
         if (exception.StatusCode == HttpStatusCode.PreconditionFailed)
         {
-            Console.WriteLine("2) Failed to update balance to 1000.0 as etags did not match.");
+            Console.WriteLine("Failed to withdraw 250 from the bank account as the local state was out of date with the remote state.");
+            Console.WriteLine("Syncing local state with remote state.");
+            currentBankAccount = await repository.GetAsync(currentBankAccount.Id);
+            Console.WriteLine($"Updated local state. {currentBankAccount}");
+
         }
     }
 
-    Console.WriteLine($"3) Forcing balance to update update balance to 1000.0 using the original etag. But explicitly telling it to ignore the etag {bankAccountUpdate}");
-    updatedBankAccount = await repository.UpdateAsync(bankAccountUpdate, ignoreEtag: true);
-    Console.WriteLine($"3) Bank account updated {updatedBankAccount}.");
-
-    BankAccount finalAccountInfo = await repository.GetAsync(createdBankAccount.Id);
-    Console.WriteLine($"Final account information {finalAccountInfo}");
-}
-
-async Task SeedAsync()
-{
-    IEnumerable<BankAccount> current = await repository.GetAsync(x => x.Type == nameof(BankAccount));
-
-    if (current.Any())
+    try
     {
-        return;
+        currentBankAccount.Withdraw(750.0);
+        Console.WriteLine("Attempting to withdraw 750 from the bank account.");
+        currentBankAccount = await repository.UpdateAsync(currentBankAccount);
+    }
+    catch (InvalidOperationException)
+    {
+        Console.WriteLine("Failed to withdraw 250 from the bank account as it would take the user overdrawn.");
     }
 
-    Faker<BankAccount> peopleFaker = new();
-    peopleFaker
-        .RuleFor(p => p.Name, f => f.Name.FullName())
-        .RuleFor(p => p.Balance, f => f.Random.Number(0, 10000));
+    currentBankAccount.Deposit(500);
+    await repository.UpdateAsync(currentBankAccount);
+    currentBankAccount.Withdraw(500);
 
-    List<BankAccount> people = peopleFaker.Generate(100);
-    await repository.CreateAsync(people);
+    Console.WriteLine($"Simulating a deposit of 500 that happened from another service. Remote State - {await repository.GetAsync(currentBankAccount.Id)}. Local State - {currentBankAccount}");
+
+    Console.WriteLine("Forcing the balance to equal 1000");
+    currentBankAccount.Balance = 1000.0;
+    currentBankAccount = await repository.UpdateAsync(currentBankAccount, ignoreEtag: true);
+
+    Console.WriteLine($"Forced update succeeded as etag was ignored.  Remote State - {await repository.GetAsync(currentBankAccount.Id)}. Local State - {currentBankAccount}");
+}
+
+async Task ConcurrencyWithOptimizeBandwidthOn()
+{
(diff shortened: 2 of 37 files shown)


Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -196,8 +210,14 @@ await container.UpsertItemAsync<TItem>(value, new PartitionKey(value.PartitionKe
 
             partitionKeyValue ??= id;
 
+            PatchItemRequestOptions patchItemRequestOptions = new();
+            if (etag != default && !string.IsNullOrWhiteSpace(etag))
```
mumby0168, line 214: Maybe we should check this further up and if it is null or whitespace intentionally then give the user a `ArgumentNullException` might be hard to detect with `default`.

Microsoft.Azure.CosmosRepository/src/IItem.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) IEvangelist. All rights reserved.
 // Licensed under the MIT License.
 
+using Newtonsoft.Json;
```
mumby0168, line 4: Need to remove this.

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -23,7 +25,18 @@ internal class InMemoryRepository<TItem> : IRepository<TItem> where TItem : clas
     {
         internal ConcurrentDictionary<string, string> Items { get; } = new();
 
-        internal string SerializeItem(TItem item) => JsonConvert.SerializeObject(item);
+        internal string SerializeItem(TItem item, string etag=null)
+        {
+            JObject jObject = JObject.FromObject(item);
+            if (etag != null)
+            {
+                jObject.Remove("_etag");
```
IEvangelist, line 33: If there is always an `"_etag"`, why not use `jObject.Replace` instead? Is there any benefit to `.Remove` and then `.Add`, or are they functionally equivalent?

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -105,23 +119,49 @@ public async ValueTask<IEnumerable<TItem>> CreateAsync(IEnumerable<TItem> values
...
+                        {
+                            throw new InvalidEtagConfigurationException(string.Empty);
+                        }
+                    }
+                    else
+                    {
+                        throw new InvalidEtagConfigurationException(string.Empty);
+                    }
+                }
+            }
```
IEvangelist, line 150:
```suggestion
            if (Items.ContainsKey(value.Id))
            {
                if (verifyEtag && value is IItemWithEtag valueWithEtag)
                {
                    TItem existingItem = DeserializeItem(Items[value.Id]);
                    if (existingItem is IItemWithEtag existingItemWithEtag)
                    {
                        if (existingItemWithEtag.Etag != valueWithEtag.Etag)
                        {
                            MismatchedEtags();
                        }
                    }
                    else
                    {
                        throw new InvalidEtagConfigurationException(string.Empty);
                    }
                }
                else
                {
                    throw new InvalidEtagConfigurationException(string.Empty);
                }
            }
```

So many nested `if` 😄 . We should work on streaming this a bit.

Microsoft.Azure.CosmosRepository/src/Item.cs
```diff
@@ -47,6 +47,15 @@ public abstract class Item : IItem
...
         public string Type { get; set; }
 
+        /// <summary>
+        /// Etag for the item which was set by Cosmos the last time the item was updated. This string is used for the relevant operations when specified.
+        /// </summary>
+        /// <remarks>
+        /// Will only be used if the verifyEtag flag is specified on the relevant methods.
+        /// </remarks>
+        [JsonProperty("_etag")]
+        public string Etag { get; set;  }
```
IEvangelist, line 57:
```suggestion
        public string Etag { get; set; }

```

Microsoft.Azure.CosmosRepository/test/Stubs/TestItemWithEtag.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosRepository;
+
+namespace Microsoft.Azure.CosmosRepositoryTests.Stubs
+{
+    public class TestItemWithEtag : Item
```
IEvangelist, line 8: Where is the ETag here?

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -155,11 +156,22 @@ await container.CreateItemAsync(value, new PartitionKey(value.PartitionKey), can
...
 
+            if (verifyEtag)
+            {
+                if (value is not IItemWithEtag valueWithEtag)
+                {
+                    throw new InvalidEtagConfigurationException($"{typeof(TItem).Name} is not assignable to {nameof(IItemWithEtag)}. You cannot specify {nameof(verifyEtag)} is true and not provide a value without an implementation of {nameof(IItemWithEtag)}.");
+                }
+
+                options.IfMatchEtag = valueWithEtag.Etag;
+            }
```
IEvangelist, line 173:
```suggestion
            if (verifyEtag && value is not IItemWithEtag valueWithEtag)
            {
                throw new InvalidEtagConfigurationException(
                    $"{typeof(TItem).Name} is not assignable to {nameof(IItemWithEtag)}. " +
                    $"You cannot specify {nameof(verifyEtag)} is true and not provide a " +
                    $"value without an implementation of {nameof(IItemWithEtag)}.");

                options.IfMatchEtag = valueWithEtag.Etag;
            }
```

Minimize nesting of `if` by using `&&`.

IEvangelist (commented):
This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?

mumby0168:
> This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?

Hi @IEvangelist 

@robertbennett1998 feel free to chip in on this, but we discussed this option, problem with it was we wanted the `Item` abstract class to be all-encompassing as in implementing all interfaces, then the client can opt-in to `etag` support if they choose. The reason we did not do this straight in was because a client may already use `Item` as their base class and all of a sudden start getting errors for the `etag` checks say if an update failed that they may not be handling.

We looked at adding another abstract class `ItemWithEtag` but it felt a little dirty if for example a user really didn't want the `etag` property on their class they could just implement the interface we provide `IItem`. Really wanted to avoid a breaking change, if we could and this seemed the nicest way to do it. We thought this would fall in line with the plans to support `LastUpdated` and `Created` times to the `Item` class we could follow the same pattern.

Hope that helps 😃

IEvangelist:
> > This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?
> 
> Hi @IEvangelist
> 
> @robertbennett1998 feel free to chip in on this, but we discussed this option, problem with it was we wanted the `Item` abstract class to be all-encompassing as in implementing all interfaces, then the client can opt-in to `etag` support if they choose. The reason we did not do this straight in was because a client may already use `Item` as their base class and all of a sudden start getting errors for the `etag` checks say if an update failed that they may not be handling.
> 
> We looked at adding another abstract class `ItemWithEtag` but it felt a little dirty if for example a user really didn't want the `etag` property on their class they could just implement the interface we provide `IItem`. Really wanted to avoid a breaking change, if we could and this seemed the nicest way to do it. We thought this would fall in line with the plans to support `LastUpdated` and `Created` times to the `Item` class we could follow the same pattern.
> 
> Hope that helps 😃

Hi @mumby0168 and @robertbennett1998 - ah, ok. I have a much better understanding now of the intention here. So, in thinking about this a bit more. I feel as though there is a bit of a tradeoff that is happening. If the idea is to eventually support `LastUpdated` and `Created` values, in addition to the ETag, the tradeoff is that (and correct me if I'm wrong):

- An all-encompassing `Item`, which includes ETag, and eventually last updated and created props. This relies on changes to the API with default parameter values to `verifyEtag`. Callers could call this with `true` on items that they want to have verified. If I'm upgrading to the latest NuGet package, and I had items in my DB would this break them to now verify the ETag? How would I know to call this with `true`, how do I opt-in, just by calling with `true`?

As compared to potentially doing something like this as an alternative:

- Leaving the `Item` class as-is, and introducing a new subclass that contains all new properties:
  - `Etag`, `LastUpdated`, and `Created`.

This could be named `FullItem`, and have all of the [system properties](https://docs.microsoft.com/en-us/azure/cosmos-db/graph/access-system-properties-gremlin). To opt-in, your items would subclass from `FullItem` instead of `Item`. It is entirely possible that existing consumers are already manually deserializing the `"_etag"` too, so as much as this isn't a breaking, it actually is. Whereas, if you're to have a separate new class - it's an explicit opt in, so existing consumers could remove their custom logic, and re-subclass to `FullItem`.

The me this is a bit less dirty, in that the API isn't changed. To opt-in you re-subclass, which is explicit, and verification happens then implicitly.

I think this would also help to avoid the future pattern work of adding many more new interfaces, `ILastUpdatedItem` and `ICreatedItem`, just thinking big picture. Thoughts on pivoting this a bit?

mumby0168:
> > > This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?
> > 
> > 
> > Hi @IEvangelist
> > @robertbennett1998 feel free to chip in on this, but we discussed this option, problem with it was we wanted the `Item` abstract class to be all-encompassing as in implementing all interfaces, then the client can opt-in to `etag` support if they choose. The reason we did not do this straight in was because a client may already use `Item` as their base class and all of a sudden start getting errors for the `etag` checks say if an update failed that they may not be handling.
> > We looked at adding another abstract class `ItemWithEtag` but it felt a little dirty if for example a user really didn't want the `etag` property on their class they could just implement the interface we provide `IItem`. Really wanted to avoid a breaking change, if we could and this seemed the nicest way to do it. We thought this would fall in line with the plans to support `LastUpdated` and `Created` times to the `Item` class we could follow the same pattern.
> > Hope that helps 😃
> 
> Hi @mumby0168 and @robertbennett1998 - ah, ok. I have a much better understanding now of the intention here. So, in thinking about this a bit more. I feel as though there is a bit of a tradeoff that is happening. If the idea is to eventually support `LastUpdated` and `Created` values, in addition to the ETag, the tradeoff is that (and correct me if I'm wrong):
> 
> * An all-encompassing `Item`, which includes ETag, and eventually last updated and created props. This relies on changes to the API with default parameter values to `verifyEtag`. Callers could call this with `true` on items that they want to have verified. If I'm upgrading to the latest NuGet package, and I had items in my DB would this break them to now verify the ETag? How would I know to call this with `true`, how do I opt-in, just by calling with `true`?
> 
> As compared to potentially doing something like this as an alternative:
> 
> * Leaving the `Item` class as-is, and introducing a new subclass that contains all new properties:
>   
>   * `Etag`, `LastUpdated`, and `Created`.
> 
> This could be named `FullItem`, and have all of the [system properties](https://docs.microsoft.com/en-us/azure/cosmos-db/graph/access-system-properties-gremlin). To opt-in, your items would subclass from `FullItem` instead of `Item`. It is entirely possible that existing consumers are already manually deserializing the `"_etag"` too, so as much as this isn't a breaking, it actually is. Whereas, if you're to have a separate new class - it's an explicit opt in, so existing consumers could remove their custom logic, and re-subclass to `FullItem`.
> 
> The me this is a bit less dirty, in that the API isn't changed. To opt-in you re-subclass, which is explicit, and verification happens then implicitly.
> 
> I think this would also help to avoid the future pattern work of adding many more new interfaces, `ILastUpdatedItem` and `ICreatedItem`, just thinking big picture. Thoughts on pivoting this a bit?

I like that idea, I still do think tho user's may have the use case to use the `etag` in some cases and not other's so I still think they're may be value in having the optional opt-out on the `UpdateAsync` method it could default to be on and it would be ignored for items that do not implement `FullItem`? 

As far as the new `FullItem` goes what do we do with the current interface `IItem` would we leave that be and create a fresh `IFullItem` interface as well?

You make a very good point tho people including my team have already created our own base type that inherits from `Item` that has created and updated time stamps, so it makes sense to create a fresh one to avoid conflicts

IEvangelist:
> > > > This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?
> > > 
> > > 
> > > Hi @IEvangelist
> > > @robertbennett1998 feel free to chip in on this, but we discussed this option, problem with it was we wanted the `Item` abstract class to be all-encompassing as in implementing all interfaces, then the client can opt-in to `etag` support if they choose. The reason we did not do this straight in was because a client may already use `Item` as their base class and all of a sudden start getting errors for the `etag` checks say if an update failed that they may not be handling.
> > > We looked at adding another abstract class `ItemWithEtag` but it felt a little dirty if for example a user really didn't want the `etag` property on their class they could just implement the interface we provide `IItem`. Really wanted to avoid a breaking change, if we could and this seemed the nicest way to do it. We thought this would fall in line with the plans to support `LastUpdated` and `Created` times to the `Item` class we could follow the same pattern.
> > > Hope that helps 😃
> > 
> > 
> > Hi @mumby0168 and @robertbennett1998 - ah, ok. I have a much better understanding now of the intention here. So, in thinking about this a bit more. I feel as though there is a bit of a tradeoff that is happening. If the idea is to eventually support `LastUpdated` and `Created` values, in addition to the ETag, the tradeoff is that (and correct me if I'm wrong):
> > 
> > * An all-encompassing `Item`, which includes ETag, and eventually last updated and created props. This relies on changes to the API with default parameter values to `verifyEtag`. Callers could call this with `true` on items that they want to have verified. If I'm upgrading to the latest NuGet package, and I had items in my DB would this break them to now verify the ETag? How would I know to call this with `true`, how do I opt-in, just by calling with `true`?
> > 
> > As compared to potentially doing something like this as an alternative:
> > 
> > * Leaving the `Item` class as-is, and introducing a new subclass that contains all new properties:
> >   
> >   * `Etag`, `LastUpdated`, and `Created`.
> > 
> > This could be named `FullItem`, and have all of the [system properties](https://docs.microsoft.com/en-us/azure/cosmos-db/graph/access-system-properties-gremlin). To opt-in, your items would subclass from `FullItem` instead of `Item`. It is entirely possible that existing consumers are already manually deserializing the `"_etag"` too, so as much as this isn't a breaking, it actually is. Whereas, if you're to have a separate new class - it's an explicit opt in, so existing consumers could remove their custom logic, and re-subclass to `FullItem`.
> > The me this is a bit less dirty, in that the API isn't changed. To opt-in you re-subclass, which is explicit, and verification happens then implicitly.
> > I think this would also help to avoid the future pattern work of adding many more new interfaces, `ILastUpdatedItem` and `ICreatedItem`, just thinking big picture. Thoughts on pivoting this a bit?
> 
> I like that idea, I still do think tho user's may have the use case to use the `etag` in some cases and not other's so I still think they're may be value in having the optional opt-out on the `UpdateAsync` method it could default to be on and it would be ignored for items that do not implement `FullItem`?
> 
> As far as the new `FullItem` goes what do we do with the current interface `IItem` would we leave that be and create a fresh `IFullItem` interface as well?
> 
> You make a very good point tho people including my team have already created our own base type that inherits from `Item` that has created and updated time stamps, so it makes sense to create a fresh one to avoid conflicts

I don't think we'd need a new interface at all, just a new class. We could still walk up to the `TItem` value and ask it if it is a `FullItem`. As for the opting in or out of ETag support, it would either you do or you don't. Right? What is the use case to not use it, in the case just inherit `Item` instead of `FullItem`.

mumby0168:
> > > > > This looks good. A general question about the API, so the consumer would have to know which items should require the additional verification of the ETag? For example, the `UpdateAsync` APIs are adding the optional parameter, but I'm curious if there is a more declarative way to achieve the same thing without impacting the consuming API. Couldn't we just check if the `TItem` is a `TItemWithEtag` and do an implicit verify?
> > > > 
> > > > 
> > > > Hi @IEvangelist
> > > > @robertbennett1998 feel free to chip in on this, but we discussed this option, problem with it was we wanted the `Item` abstract class to be all-encompassing as in implementing all interfaces, then the client can opt-in to `etag` support if they choose. The reason we did not do this straight in was because a client may already use `Item` as their base class and all of a sudden start getting errors for the `etag` checks say if an update failed that they may not be handling.
> > > > We looked at adding another abstract class `ItemWithEtag` but it felt a little dirty if for example a user really didn't want the `etag` property on their class they could just implement the interface we provide `IItem`. Really wanted to avoid a breaking change, if we could and this seemed the nicest way to do it. We thought this would fall in line with the plans to support `LastUpdated` and `Created` times to the `Item` class we could follow the same pattern.
> > > > Hope that helps 😃
> > > 
> > > 
> > > Hi @mumby0168 and @robertbennett1998 - ah, ok. I have a much better understanding now of the intention here. So, in thinking about this a bit more. I feel as though there is a bit of a tradeoff that is happening. If the idea is to eventually support `LastUpdated` and `Created` values, in addition to the ETag, the tradeoff is that (and correct me if I'm wrong):
> > > 
> > > * An all-encompassing `Item`, which includes ETag, and eventually last updated and created props. This relies on changes to the API with default parameter values to `verifyEtag`. Callers could call this with `true` on items that they want to have verified. If I'm upgrading to the latest NuGet package, and I had items in my DB would this break them to now verify the ETag? How would I know to call this with `true`, how do I opt-in, just by calling with `true`?
> > > 
> > > As compared to potentially doing something like this as an alternative:
> > > 
> > > * Leaving the `Item` class as-is, and introducing a new subclass that contains all new properties:
> > >   
> > >   * `Etag`, `LastUpdated`, and `Created`.
> > > 
> > > This could be named `FullItem`, and have all of the [system properties](https://docs.microsoft.com/en-us/azure/cosmos-db/graph/access-system-properties-gremlin). To opt-in, your items would subclass from `FullItem` instead of `Item`. It is entirely possible that existing consumers are already manually deserializing the `"_etag"` too, so as much as this isn't a breaking, it actually is. Whereas, if you're to have a separate new class - it's an explicit opt in, so existing consumers could remove their custom logic, and re-subclass to `FullItem`.
> > > The me this is a bit less dirty, in that the API isn't changed. To opt-in you re-subclass, which is explicit, and verification happens then implicitly.
> > > I think this would also help to avoid the future pattern work of adding many more new interfaces, `ILastUpdatedItem` and `ICreatedItem`, just thinking big picture. Thoughts on pivoting this a bit?
> > 
> > 
> > I like that idea, I still do think tho user's may have the use case to use the `etag` in some cases and not other's so I still think they're may be value in having the optional opt-out on the `UpdateAsync` method it could default to be on and it would be ignored for items that do not implement `FullItem`?
> > As far as the new `FullItem` goes what do we do with the current interface `IItem` would we leave that be and create a fresh `IFullItem` interface as well?
> > You make a very good point tho people including my team have already created our own base type that inherits from `Item` that has created and updated time stamps, so it makes sense to create a fresh one to avoid conflicts
> 
> I don't think we'd need a new interface at all, just a new class. We could still walk up to the `TItem` value and ask it if it is a `FullItem`. As for the opting in or out of ETag support, it would either you do or you don't. Right? What is the use case to not use it, in the case just inherit `Item` instead of `FullItem`.

Okay makes sense for the interface, I think the case for the `FullItem` be used without `etags` would be when they want the `created` and `updated` stamps but don't want `etag` support, what do you think? I kinda like the option to override the "norm" via the method

IEvangelist:
Starting a new thread, as the quote reply was getting a bit long. @mumby0168 - is the last updated and created dates custom, or already part of the internal system provided values? Also, the interface does make sense if we want to restrict the consumer from setting it. Right?

mumby0168:
> Starting a new thread, as the quote reply was getting a bit long. @mumby0168 - is the last updated and created dates custom, or already part of the internal system provided values? Also, the interface does make sense if we want to restrict the consumer from setting it. Right?
> 
> Found this too: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L791-L886

Good idea lol, the `lastUpdated` could be inferred from the `ts` property and converted to a `DateTime`, `created` maybe not so much see #78 think we discussed this before

IEvangelist:
> > Starting a new thread, as the quote reply was getting a bit long. @mumby0168 - is the last updated and created dates custom, or already part of the internal system provided values? Also, the interface does make sense if we want to restrict the consumer from setting it. Right?
> > Found this too: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L791-L886
> 
> Good idea lol, the `lastUpdated` could be inferred from the `ts` property and converted to a `DateTime`, `created` maybe not so much see #78 think we discussed this before

So really, `lastUpdated` we get for free, just need to parse it/map it out. As for the `created` that would need to be set in the create calls or something, when configured? So maybe, two new classes? `DatedItem`, and `ETagItem`. I see where you're getting at, it seems to grow either way. What is your preference?

mumby0168:
> > > Starting a new thread, as the quote reply was getting a bit long. @mumby0168 - is the last updated and created dates custom, or already part of the internal system provided values? Also, the interface does make sense if we want to restrict the consumer from setting it. Right?
> > > Found this too: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L791-L886
> > 
> > 
> > Good idea lol, the `lastUpdated` could be inferred from the `ts` property and converted to a `DateTime`, `created` maybe not so much see #78 think we discussed this before
> 
> So really, `lastUpdated` we get for free, just need to parse it/map it out. As for the `created` that would need to be set in the create calls or something, when configured? So maybe, two new classes? `DatedItem`, and `ETagItem`. I see where you're getting at, it seems to grow either way. What is your preference?

Yeah that is what I was thinking, we can possibly still use the `FullItem` which just have everything or we can offer the subsets like you are proposing with `DatedItem` & `ETagItem` so if a user wants to derive their own type via a hierarchy they can do, in my opinion the more flexible we make it the better, what do you think?

robertbennett1998:
> > > > Starting a new thread, as the quote reply was getting a bit long. @mumby0168 - is the last updated and created dates custom, or already part of the internal system provided values? Also, the interface does make sense if we want to restrict the consumer from setting it. Right?
> > > > Found this too: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos.Samples/Usage/ItemManagement/Program.cs#L791-L886
> > > 
> > > 
> > > Good idea lol, the `lastUpdated` could be inferred from the `ts` property and converted to a `DateTime`, `created` maybe not so much see #78 think we discussed this before
> > 
> > 
> > So really, `lastUpdated` we get for free, just need to parse it/map it out. As for the `created` that would need to be set in the create calls or something, when configured? So maybe, two new classes? `DatedItem`, and `ETagItem`. I see where you're getting at, it seems to grow either way. What is your preference?
> 
> Yeah that is what I was thinking, we can possibly still use the `FullItem` which just have everything or we can offer the subsets like you are proposing with `DatedItem` & `ETagItem` so if a user wants to derive their own type via a hierarchy they can do, in my opinion the more flexible we make it the better, what do you think?

I think changing it to leave Item as it is and introducing FullItem works, but I think we should have interfaces following the same pattern as IItemWithETag as It would remove the vertical hierarchy of multiple abstract classes and allow a flat hierarchy. 

Where do we stand on the verifyETag property in the update methods? I personally still think this is a good idea as it allows whether or not to override it for a given operation. However, I do see your point about if you are using the FullItem type you probably want this to happen implicitly. What about flipping it and having an ignoreETag flag instead?

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
robertbennett1998: I have updated to use:
`jObject["_etag"] = JToken.FromObject(etag);`

Microsoft.Azure.CosmosRepository/test/Stubs/TestItemWithEtag.cs
robertbennett1998: It is currently in Item, I will be updating in line with the discussion below.

mumby0168:
Hey @all-contributors please add @robertbennett1998 for tests and code

Microsoft.Azure.CosmosRepository/src/EtagItem.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Microsoft.Azure.CosmosRepository
+{
+    /// <summary>
+    /// A base helper class that implements IItemWithEtag
+    /// </summary>
+    /// <example>
+    /// Here is an example subclass item, which adds several properties:
+    /// <code language="c#">
+    /// <![CDATA[
+    /// public class SubItem : FullItem
```
mumby0168:
```suggestion
    /// public class SubItem : EtagItem
```

Microsoft.Azure.CosmosRepository/src/EtagItem.cs
```diff
@@ -0,0 +1,79 @@
...
+    ///
+    /// public class Child
+    /// {
+    ///     public string Name { get; set; }
+    ///     public DateTime BirthDate { get; set; }
+    /// }
+    /// ]]>
+    /// </code>
+    /// </example>
+    public class EtagItem : IItemWithEtag
```
mumby0168, line 33: Can this not derive from `Item`?

Microsoft.Azure.CosmosRepository/src/FullItem.cs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// public class Child
+    /// {
+    ///     public string Name { get; set; }
+    ///     public DateTime BirthDate { get; set; }
+    /// }
+    /// ]]>
+    /// </code>
+    /// </example>
+    public abstract class FullItem : IItemWithEtag
```
mumby0168, line 36: Again as well as implementing the interface it could also derive from `Item` save writing out the properties again

Microsoft.Azure.CosmosRepository/src/EtagItem.cs
robertbennett1998: That is a good point :D

Microsoft.Azure.CosmosRepository/src/EtagItem.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class EtagItem : Item, IItemWithEtag
+    {
+        /// <summary>
+        /// Etag for the item which was set by Cosmos the last time the item was updated. This string is used for the relevant operations when specified.
+        /// </summary>
+        /// <remarks>
+        /// Will only be used if the verifyEtag flag is specified on the relevant methods.
+        /// </remarks>
+        [JsonProperty("_etag")]
+        public string Etag { get; set; }
```
mumby0168, line 42: Do you think there is a way to stop the user from setting this value, while still being able to de-serialize it?

robertbennett1998: Good question I will have a look into it

robertbennett1998: Using the JsonProperty attribute should be enough. 
https://www.newtonsoft.com/json/help/html/serializationattributes.htm#JsonPropertyAttribute

mumby0168: Sounds good 👍

Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using OptimisticConcurrencyControl;
+
+ConfigurationBuilder configuration = new();
+
+ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
+    {
+        options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+        options.DatabaseId = "optimistic-concurrency-control";
+        options.OptimizeBandwidth = false; // Must be false to receive the upto date etags back on update calls
```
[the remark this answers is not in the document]
robertbennett1998: Yeah, I have rewritten the examples to highlight this better

Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/Program.cs
```diff
@@ -0,0 +1,208 @@
...
+    Console.WriteLine($"Simulating a withdrawal of 500 that happened from another service. Remote State - {await repository.GetAsync(currentBankAccount.Id)}. Local State - {currentBankAccount}");
+
+    Console.WriteLine("Attempting to withdraw 250 from the bank account.");
+    try
+    {
+        Console.WriteLine("Attempting to withdraw 250 from the bank account.");
+        await repository.UpdateAsync(currentBankAccount.Id,
+            builder => builder.Replace(account => account.Balance, currentBankAccount.Balance - 250), etag: currentBankAccount.Etag);
+    }
+    catch (CosmosException exception)
```
mumby0168, line 62:
```suggestion
    catch (CosmosException exception) when (excepton.StatusCode = StatusCode)
```

Then you can drop the nested `if`

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -124,20 +124,24 @@ public interface IRepository<TItem> where TItem : IItem
         /// </summary>
         /// <param name="value">The item value to update.</param>
         /// <param name="cancellationToken">The cancellation token to use when making asynchronous operations.</param>
+        /// <param name="ignoreEtag">When TItem implements IItemWithEtag the etag will be verified on all updates. Setting this flag to true indicates that the</param>
```
mumby0168, line 127: Think the sentence just needs finishing off here

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -124,20 +124,24 @@ public interface IRepository<TItem> where TItem : IItem
...
-            CancellationToken cancellationToken = default);
+            CancellationToken cancellationToken = default,
+            bool ignoreEtag = false);
 
         /// <summary>
         /// Updates one or more cosmos item(s) representing the given <paramref name="values"/>.
         /// </summary>
         /// <param name="values">The item values to update.</param>
         /// <param name="cancellationToken">The cancellation token to use when making asynchronous operations.</param>
+        /// <param name="ignoreEtag">When TItem implements IItemWithEtag the etag will be verified on all updates. Setting this flag to true indicates that the</param>
```
mumby0168, line 139: Same here just needs finishing off.

Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
```diff
@@ -0,0 +1,20 @@
...
+        <UserSecretsId>fd4da918-55af-4010-9ba0-e75f0e0dc917</UserSecretsId>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\Microsoft.Azure.CosmosRepository\src\Microsoft.Azure.CosmosRepository.csproj" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Bogus" Version="33.0.2" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="5.0.0" />
```
IEvangelist, line 17:
```suggestion
      <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="6.0.0" />
```

Microsoft.Azure.CosmosRepository/test/Stubs/TestItemWithEtag.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+    /// <summary>
+    /// Etag for the item which was set by Cosmos the last time the item was updated. This string is used for the relevant operations when specified.
+    /// </summary>
+    /// <remarks>
+    /// Will only be used if the verifyEtag flag is specified on the relevant methods.
+    /// </remarks>
+    [JsonProperty("_etag")]
+    public string Etag { get; set; }
+
```
IEvangelist, line 19:
```suggestion
        /// <summary>
        /// Etag for the item which was set by Cosmos the last time the item was updated. This string is used for the relevant operations when specified.
        /// </summary>
        /// <remarks>
        /// Will only be used if the verifyEtag flag is specified on the relevant methods.
        /// </remarks>
        [JsonProperty("_etag")]
        public string Etag { get; set; }
```

Microsoft.Azure.CosmosRepository/src/IItem.cs
```diff
@@ -23,4 +23,5 @@ public interface IItem
         /// </summary>
         string PartitionKey { get; }
     }
+
```
IEvangelist, line 26:
```suggestion
```

Microsoft.Azure.CosmosRepository/src/FullItem.cs
```diff
@@ -0,0 +1,45 @@
...
+    ///
+    /// public class Child
+    /// {
+    ///     public string Name { get; set; }
+    ///     public DateTime BirthDate { get; set; }
+    /// }
+    /// ]]>
+    /// </code>
+    /// </example>
+    public abstract class FullItem : Item, IItemWithEtag
```
IEvangelist, line 34: Why is this `abstract` and what's the difference between this and the `EtagItem`? Why not inherit from `EtagItem` here, and add the two date properties?

IEvangelist (approved):
Really close to done, a few nits, and questions.

Microsoft.Azure.CosmosRepository/src/FullItem.cs
mumby0168: Does it depend on what the plan is I suppose do we want the time stamp changes in this PR? I thought we could do that as a separate piece of work, then we can have the `DatedItem` as an option for time stamps without the `ETagItem` then the `FullItem` could include all of `Item`, `DatedItem` & `ETagItem` 😃  it a lot of base types I know but it offers the most flexibility from a user point of view, what do you both think? Makes sense that this would not strictly *have* to be `abstract` as well I suppose.

Microsoft.Azure.CosmosRepository.Samples/OptimisticConcurrencyControl/OptimisticConcurrencyControl.csproj
robertbennett1998: I will remove this as secrets isn't actually been used now.

Microsoft.Azure.CosmosRepository/src/FullItem.cs
IEvangelist: Sounds good to me, thanks for the clarification.

mumby0168: As for the `abstract` we had a chat and did not see a use case for wanted an `IRepository<FullItem>` just on it's own with nothing else so we chose to leave it as is, will hang fire with the release just confirm :)

mumby0168 (approved):
Great work @robertbennett1998

PR #156, opened 2022-01-04
Fix error being thrown when additional attributes are applied to poco classes

WestDiscGolf:
Updated the attribute checking as when I was testing with my POCO it had additional attributes, in my example [Required] for validation, and it was failing on the update.

### DIFF
--- a/Microsoft.Azure.CosmosRepository/src/Builders/PatchOperationBuilder.cs
+++ b/Microsoft.Azure.CosmosRepository/src/Builders/PatchOperationBuilder.cs
@@ -43,8 +43,7 @@ namespace Microsoft.Azure.CosmosRepository.Builders
 
             foreach (object attribute in attributes)
             {
-                JsonPropertyAttribute jsonAttribute = (JsonPropertyAttribute) attribute;
-                if (jsonAttribute is not null)
+                if (attribute is JsonPropertyAttribute jsonAttribute)
                 {
                     return jsonAttribute.PropertyName;
                 }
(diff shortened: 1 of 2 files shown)


Microsoft.Azure.CosmosRepository/src/Builders/PatchOperationBuilder.cs
```diff
@@ -43,8 +43,7 @@ private string GetPropertyToReplace(PropertyInfo propertyInfo)
 
             foreach (object attribute in attributes)
             {
-                JsonPropertyAttribute jsonAttribute = (JsonPropertyAttribute) attribute;
-                if (jsonAttribute is not null)
+                if (attribute is JsonPropertyAttribute jsonAttribute)
```
IEvangelist, line 46: Love this 👍🏼

mumby0168, line 46: Was this throwing an exception here I imagine? Since it was using a direct case rather than `as`?

IEvangelist (approved):

mumby0168 (approved):
Thanks for this PR, looks good to me 😃

Just one for the future this PR is okay with it being so small but it is usually best to start by raising an issue.

Thanks,
Billy

IEvangelist:
@all-contributors please add @WestDiscGolf for code and tests

Microsoft.Azure.CosmosRepository/src/Builders/PatchOperationBuilder.cs
WestDiscGolf: Indeed 😄

PR #159, opened 2022-01-05
Adding CountAsync Methods

### LINKED ISSUE #84: Implement CountAsync<TItem>
I would like to request a count of records for a given filter criteria/query. For example:
`SELECT VALUE COUNT(1) FROM c WHERE c....`

When querying CosmosDb directly, this returns a response as follows:
```
[
    27
]
``` 
(In the example above, there are 27 records)

Using the repository, this doesn't seem possible. Is there a way to use the existing repository to get the above result back, or would a new method be required?

Since all POCOs must derive from IItem, I have tried to create a POCO for the anonymous array response type above but have faced some challenges (with deserialization, etc). I don't this its possible (but could be wrong, only spent a few minutes experimenting).

kevin-benton:
Resolves #84 

Adding 2 new methods to the `Repository` interface.
1. A `CountAsync(CancellationToken)` method that returns the count of all items in the container.
2. A `CountAsync(Expression<Func<TItem, bool>>, CancellationToken)` method that returns the count of the items matching the query.

Tests were added for the `InMemoryRepository`.

I'm open to adding examples to the example projects but have not found a great place to put them yet.


Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -249,6 +249,21 @@ public async ValueTask<bool> ExistsAsync(Expression<Func<TItem, bool>> predicate
                 item => item.Type == typeof(TItem).Name, Expression.AndAlso).Compile());
         }
 
+        /// <inheritdoc/>
+        public async ValueTask<int> CountAsync(CancellationToken cancellationToken = default)
+        {
+            await Task.CompletedTask;
+            return Items.Values.Count;
```
mumby0168, line 256: Should this be checking the type of the items count, say you have multiple `TItem`s within the same container? Depends on the intention of the API I think.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
```diff
@@ -227,6 +227,29 @@ public interface IRepository<TItem> where TItem : IItem
...
 
+        /// <summary>
+        /// Queries cosmos DB to obtain the count of items.
+        /// </summary>
+        /// <remarks>
+        /// This queries the total number of documents in the container.
+        /// </remarks>
+        /// <param name="cancellationToken">The cancellation token to use when making asynchronous operations.</param>
+        /// <returns>A <see cref="ValueTask"/> representing the asynchronous count operation.</returns>
+        ValueTask<int> CountAsync(CancellationToken cancellationToken = default);
```
mumby0168, line 238: What was your intention behind this API, count of the documents in the container of the given type? Just wondering if we need this or the below API could be made to do this by the user of the library if that makes sense?

mumby0168:
Looks a great start just a few questions on one of the API's

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
kevin-benton: You're correct. I will make the update. In the `DefaultRepository` it filters by type.

kevin-benton: I've added the filter in the implementation.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
kevin-benton: I went both ways on this. I spent a lot of time looking at my personal work and when I needed the count of a certain type. Most of the times, I used a filter to get a count of certain documents, but sometimes I did want the total count of one document type.

But yes this all could be achieved with the expression API such as - `CountAsync(item => true)` or similar. However, it also seems to me that code would be messier than `CountAsync()` in someone's implementation. That's why I left it in.

But now that I'm thinking about it, we could make an extension method to provide a no predicate option. I'm fine with any of the 3 options: leave as is, remove the call without the predicate, or add an extension method that hides the predicate.

mumby0168: I like extension methods my only hesitation is testability as you can't mock extension methods. I think since each `IRepository<T>` is scoped to the specific `IItem` then it makes sense that this would only count those `IItem`s in a container. 

What do you think @IEvangelist ?

IEvangelist (commented):
A few nits, otherwise this looks great.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
kevin-benton: So I believe we are saying to leave both methods?

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -303,6 +303,33 @@ await container.ReadItemAsync<TItem>(id, partitionKey, cancellationToken: cancel
...
+                await _containerProvider.GetContainerAsync().ConfigureAwait(false);
+
+            IQueryable<TItem> query = container.GetItemLinqQueryable<TItem>();
+
+            return await _cosmosQueryableProcessor.CountAsync(query, cancellationToken);
+        }
+
+        /// <inheritdoc/>
+        public async ValueTask<int> CountAsync(Expression<Func<TItem, bool>> predicate,
+            CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
kevin-benton: Updated with the suggested change.

Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs
```diff
@@ -303,6 +303,33 @@ await container.ReadItemAsync<TItem>(id, partitionKey, cancellationToken: cancel
...
+        {
+            Container container =
+                await _containerProvider.GetContainerAsync().ConfigureAwait(false);
+
+            IQueryable<TItem> query =
+                container.GetItemLinqQueryable<TItem>()
+                    .Where(_repositoryExpressionProvider.Build(predicate));
+
+            TryLogDebugDetails(_logger, () => $"Read: {query}");
```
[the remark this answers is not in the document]
kevin-benton: Updated with conditional logging in both implementations.

Microsoft.Azure.CosmosRepository/src/IRepository.cs
mumby0168: Yes I think so

IEvangelist:
@all-contributors please add @kevin-benton for code and tests

mumby0168 (approved):

PR #163, opened 2022-01-08
Change Feed Support

### LINKED ISSUE #162: Change Feed Processor Support
The SDK that we are currently wrapping with this library also includes the code to allow the change feed to be processed for a given container. It would be nice if this library could expose an easy way to listen from the change feed for a given container. This could use our strongly typed notion of `IItem`s to allow an interface such as below to be defined.

```csharp
using Microsoft.Azure.CosmosRepository;

public interface IChangeFeedItemProcessor<in TItem> where TItem : IItem
{
    ValueTask HandleAsync(T item, CancellationToken cancellationToken);
}
```

The library could be set up to listen for changes in the change feed for a given container and then for only given `IItem`s within that container.

I think this would follow the goals and intentions of the library which is to make using cosmos super easy, in this case all a client would need to do is say they want to listen to the change feed at startup, then once the app has started a `HostedService` could be running and calling the correct `IChangeFeedItemProcessor<in TItem>` every time a change is detected.

See the example below

```csharp
using Microsoft.Azure.CosmosRepository;
using Microsoft.Extensions.Logging;

public class TestItem : Item
{
    
}

public class TestItemChangeFeedItemProcessor : IChangeFeedItemProcessor<TestItem>
{
    private readonly ILogger<TestItemChangeFeedItemProcessor> _logger;

    public TestItemChangeFeedItemProcessor(ILogger<TestItemChangeFeedItemProcessor> logger) => 
        _logger = logger;

    public ValueTask HandleAsync(TestItem item, CancellationToken cancellationToken)
    {
        _logger.LogInformation("Change detected on item with ID {ItemId}", item.Id);
        return ValueTask.CompletedTask;
    }
}
```

## In-Memory Support

This implementation could also have 1st class in memory support, we could in the in-memory repository try and invoke the correct `IChangeFeedItemProcessor<TItem>` for any creates/updates performed in-memory.

mumby0168:
Closes #162


mumby0168:
Hi, @IEvangelist sorry for the large PR I am going to leave in-memory support for the change feed as part of a separate PR, conscious of the size.

I have left a few `//TODO:`s in with a few questions that I think we could do with discussing, certainly not finished yet, I have a lot of tests to add but I wanted to see your opinion on the intention and the initial implementation of the idea before going back and adding the finishes touches and tests 😃

mumby0168:
See this PR for how the in-memory change feed is used in test

https://github.com/mumby0168/Nomad.Sorter/pull/3

Microsoft.Azure.CosmosRepository/src/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+            if(!typeof(IItem).IsAssignableFrom(type))
+            {
+                throw new InvalidOperationException(
+                    $"The type {type.FullName} does not implement {typeof(IItem).FullName}");
+            }
+        }
+
+        public static void EnsureAllAreTypeOfIItem(this IReadOnlyList<Type> types)
+        {
+            foreach (Type type in types) type.EnsureIsTypeOfIItem();
```
IEvangelist, line 22:
```suggestion
            foreach (Type type in types) type.IsIItem();
```

Microsoft.Azure.CosmosRepository/src/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static void EnsureIsTypeOfIItem(this Type type)
+        {
+            if(!typeof(IItem).IsAssignableFrom(type))
+            {
+                throw new InvalidOperationException(
+                    $"The type {type.FullName} does not implement {typeof(IItem).FullName}");
+            }
+        }
+
+        public static void EnsureAllAreTypeOfIItem(this IReadOnlyList<Type> types)
```
IEvangelist, line 20:
```suggestion
        public static void AllAreIItems(this IReadOnlyList<Type> types)
```

Not sure I like this name either :)

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
```diff
@@ -27,6 +28,8 @@ internal class InMemoryRepository<TItem> : IRepository<TItem> where TItem : clas
         internal long CurrentTs => DateTimeOffset.UtcNow.ToUnixTimeSeconds();
         internal ConcurrentDictionary<string, string> Items { get; } = new();
 
+        internal EventHandler<ChangeFeedItemArgs<TItem>> Changes { get; set; }
```
IEvangelist, line 31: Could this just be an `Action<T>` instead an `EventHandler<T>`?

Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
```diff
@@ -51,6 +51,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Scrutor" Version="3.3.0" />
```
IEvangelist, line 54: This additional dependency is a bit concerning in that it adds a new dependency for something that is only optional. In other words, even if a user is not using change feeds they'll get this package. Might need to think this over a bit more.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerSyncContainerPropertiesProvider.cs
```diff
@@ -1,12 +1,14 @@
...
 
+using System;
+
 namespace Microsoft.Azure.CosmosRepository.Providers
 {
     /// <summary>
     /// Decides whether or not a given container should sync it's properties.
     /// </summary>
-    public interface ICosmosContainerSyncContainerPropertiesProvider
+    interface ICosmosContainerSyncContainerPropertiesProvider
```
IEvangelist, line 11: Is this a breaking change?

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerDefaultTimeToLiveProvider.cs
```diff
@@ -1,12 +1,14 @@
...
 
+using System;
+
 namespace Microsoft.Azure.CosmosRepository.Providers
 {
     /// <summary>
     /// The cosmos container default time to live provider to an <see cref="IItem"/>
     /// </summary>
-    public interface ICosmosContainerDefaultTimeToLiveProvider
+    interface ICosmosContainerDefaultTimeToLiveProvider
```
IEvangelist, line 11: Is this a breaking change?

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosThroughputProvider.cs
```diff
@@ -1,20 +1,23 @@
...
+using System;
 using Microsoft.Azure.Cosmos;
 
 namespace Microsoft.Azure.CosmosRepository.Providers
 {
     /// <summary>
     /// Exposes the ability to get an <see cref="IItem"/>s containers throughput properties.
     /// </summary>
-    public interface ICosmosThroughputProvider
+    interface ICosmosThroughputProvider
```
IEvangelist, line 12: Is this a breaking change?

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerSyncContainerPropertiesProvider.cs
```diff
@@ -15,5 +17,7 @@ public interface ICosmosContainerSyncContainerPropertiesProvider
         /// <remarks>If the SyncAllContainerProperties is set to true this will override any specific containers configuration.</remarks>
         /// <returns>Whether or not to sync container properties</returns>
         bool GetWhetherToSyncContainerProperties<TItem>() where TItem : IItem;
```
IEvangelist, line 19: Might be a good use case for default interface members.
```suggestion
        bool GetWhetherToSyncContainerProperties<TItem>() where TItem : IItem =>
            GetWhetherToSyncContainerProperties(typeof(TItem));
```

Microsoft.Azure.CosmosRepository/src/Services/DefaultCosmosContainerService.cs
```diff
@@ -78,5 +83,23 @@ class DefaultCosmosContainerService : ICosmosContainerService
...
+        public Task<Container> GetContainerAsync(IReadOnlyList<Type> itemTypes)
+        {
+            if (itemTypes.Any() is false)
+            {
+                throw new InvalidOperationException("You must provided at least one item type to get a container for");
+            }
+
+            string containerName = _cosmosItemConfigurationProvider.GetOptions(itemTypes.First()).ContainerName;
+
+            if(itemTypes.Select(x => _cosmosItemConfigurationProvider.GetOptions(x)).All(x => x.ContainerName == containerName))
```
IEvangelist, line 96:
```suggestion
            if (itemTypes.Select(x => _cosmosItemConfigurationProvider.GetOptions(x)).All(x => x.ContainerName == containerName))
```

IEvangelist (commented):
A few nits on styling, a few questions, a few suggestions. I love the way this is shaping up. Curious about maybe splitting this into a separate package, similar to what I did here with:

- https://www.nuget.org/packages/HaveIBeenPwned.Client
- https://www.nuget.org/packages/HaveIBeenPwned.Client.PollyExtensions

They share the same repo and everything. For example, we might have something like `IEvangelist.Azure.CosmosRepository.ChangeFeed`. It would be a stand alone, and it would have all this same stuff. It would allow for an opt-in, and alleviate the concern about the new NuGet dependency. Thoughts or concerns on that?

mumby0168:
> A few nits on styling, a few questions, a few suggestions. I love the way this is shaping up. Curious about maybe splitting this into a separate package, similar to what I did here with:
> 
> * https://www.nuget.org/packages/HaveIBeenPwned.Client
> * https://www.nuget.org/packages/HaveIBeenPwned.Client.PollyExtensions
> 
> They share the same repo and everything. For example, we might have something like `IEvangelist.Azure.CosmosRepository.ChangeFeed`. It would be a stand-alone, and it would have all this same stuff. It would allow for an opt-in, and alleviate the concern about the new NuGet dependency. Thoughts or concerns on that?

I did consider the separate package, my thoughts were that since the SDK that we are wrapping also has the changed feed we may as well follow the same pattern, which then means you already have this capability all you have to do is opt-in? I am happy to drop the `Scrutor` package it just makes assembly scanning easier but I can for sure do that without the need for the extra dependency does that suit?

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerSyncContainerPropertiesProvider.cs
mumby0168: I like this.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosThroughputProvider.cs
mumby0168: Potentially yes, these should not really be external however, they seem to have slipped through the net, what is your thought's on this what should we do?

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
mumby0168: We could if we are sure there are no more subscribers to this than the one we have currently, does it hurt to leave it as it?

Microsoft.Azure.CosmosRepository/src/Microsoft.Azure.CosmosRepository.csproj
mumby0168: I can do this assembly scanning without taking on the package does that suit?

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerSyncContainerPropertiesProvider.cs
mumby0168: @IEvangelist this can be done but only if we bump the package up to `netstandard2.1` which would mean we drop .NET framework support, what are your thoughts on this?

IEvangelist: Yeah, it's not worth it. Using a somewhat cool feature vs. dropping support for an entire framework. Let's not do that 😄

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosThroughputProvider.cs
IEvangelist: I'm fine with it, probably wasn't used externally.

IEvangelist:
> > A few nits on styling, a few questions, a few suggestions. I love the way this is shaping up. Curious about maybe splitting this into a separate package, similar to what I did here with:
> > 
> > * https://www.nuget.org/packages/HaveIBeenPwned.Client
> > * https://www.nuget.org/packages/HaveIBeenPwned.Client.PollyExtensions
> > 
> > They share the same repo and everything. For example, we might have something like `IEvangelist.Azure.CosmosRepository.ChangeFeed`. It would be a stand-alone, and it would have all this same stuff. It would allow for an opt-in, and alleviate the concern about the new NuGet dependency. Thoughts or concerns on that?
> 
> I did consider the separate package, my thoughts were that since the SDK that we are wrapping also has the changed feed we may as well follow the same pattern, which then means you already have this capability all you have to do is opt-in? I am happy to drop the `Scrutor` package it just makes assembly scanning easier but I can for sure do that without the need for the extra dependency does that suit?
> 
> However, it may be useful to create a package may be called `IEvangelist.Microsoft.Azure.CosmosRepository.AspNetCore` that includes the hosted service I created in the sample app to process the change feed, this could then include `Scrutor` for the assembly scanning what do you think? See the hosted service here: https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/users/wmumby/changeFeedSupport/Microsoft.Azure.CosmosRepository.Samples/ChangeFeedWeb/Services/ChangeFeedHostedService.cs

I love that idea! So, we'd drop `Scrutor` for the main package - but then expose a new package with it that has the hosted service bits? Right? If so, let's do that.

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosThroughputProvider.cs
mumby0168: Okay cool we'll leave it as is 😃

Microsoft.Azure.CosmosRepository/src/Providers/ICosmosContainerSyncContainerPropertiesProvider.cs
mumby0168: Cool, I thought the same

mumby0168:
> > > A few nits on styling, a few questions, a few suggestions. I love the way this is shaping up. Curious about maybe splitting this into a separate package, similar to what I did here with:
> > > 
> > > * https://www.nuget.org/packages/HaveIBeenPwned.Client
> > > * https://www.nuget.org/packages/HaveIBeenPwned.Client.PollyExtensions
> > > 
> > > They share the same repo and everything. For example, we might have something like `IEvangelist.Azure.CosmosRepository.ChangeFeed`. It would be a stand-alone, and it would have all this same stuff. It would allow for an opt-in, and alleviate the concern about the new NuGet dependency. Thoughts or concerns on that?
> > 
> > 
> > I did consider the separate package, my thoughts were that since the SDK that we are wrapping also has the changed feed we may as well follow the same pattern, which then means you already have this capability all you have to do is opt-in? I am happy to drop the `Scrutor` package it just makes assembly scanning easier but I can for sure do that without the need for the extra dependency does that suit?
> > However, it may be useful to create a package may be called `IEvangelist.Microsoft.Azure.CosmosRepository.AspNetCore` that includes the hosted service I created in the sample app to process the change feed, this could then include `Scrutor` for the assembly scanning what do you think? See the hosted service here: https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/users/wmumby/changeFeedSupport/Microsoft.Azure.CosmosRepository.Samples/ChangeFeedWeb/Services/ChangeFeedHostedService.cs
> 
> I love that idea! So, we'd drop `Scrutor` for the main package - but then expose a new package with it that has the hosted service bits? Right? If so, let's do that.

How would you want this to look in the package? Another directory called `Microsoft.Azure.CosmosRepository.AspNetCore` with a `src` and `tests` folder?

Microsoft.Azure.CosmosRepository/src/ChangeFeed/Providers/DefaultChangeFeedContainerProcessorProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+            IEnumerable<IGrouping<string, ContainerOptionsBuilder>> containers = repositoryOptions.ContainerBuilder
+                .Options
+                .Where(x => x.ChangeFeedOptions != null)
+                .GroupBy(x => x.Name);
+
+            foreach (IGrouping<string, ContainerOptionsBuilder> container in containers)
+            {
+                List<Type> itemTypes = container.Select(x => x.Type).ToList();
+
+                yield return new DefaultContainerChangeFeedProcessor(
```
mumby0168, line 53: @robertbennett1998 could this be tidied up a little?

mumby0168: I was thinking rather than passing the services down, create the two containers here and pass them in? Makes this class more standalone and not tied to DI and injecting these providers

IEvangelist:
> How would you want this to look in the package? Another directory called `Microsoft.Azure.CosmosRepository.AspNetCore` with a `src` and `tests` folder?
> 
> Or shall we re-organize a little and have the root of the repo have `src` `tests` and `samples` folders to store the projects we currently have?

I think we should re-org everything, just have top level src, tests, and samples. That will be a large bit of changes, let's try to isolate them to a single commit and review that single commit for ease-of-reviewability.

mumby0168:
> > How would you want this to look in the package? Another directory called `Microsoft.Azure.CosmosRepository.AspNetCore` with a `src` and `tests` folder?
> > Or shall we re-organize a little and have the root of the repo have `src` `tests` and `samples` folders to store the projects we currently have?
> 
> I think we should re-org everything, just have top level src, tests, and samples. That will be a large bit of changes, let's try to isolate them to a single commit and review that single commit for ease-of-reviewability.

Should we look at getting the code we have up to scratch merge this PR and then look at re-organizing the repo as another PR? Then we can add the extra package we have been talking of as one more pR?

IEvangelist:
> Should we look at getting the code we have up to scratch merge this PR and then look at re-organizing the repo as another PR? Then we can add the extra package we have been talking of as one more pR?

Honestly, I'm good either way. Whichever way you think is the easiest - I'm good with

mumby0168:
> > Should we look at getting the code we have up to scratch merge this PR and then look at re-organizing the repo as another PR? Then we can add the extra package we have been talking of as one more pR?
> 
> Honestly, I'm good either way. Whichever way you think is the easiest - I'm good with

If it's okay with you shall we review this and merge it? Then I'll do another PR for the re-structure, if not this PR would be very bloated 😃

Microsoft.Azure.CosmosRepository/src/ChangeFeed/IContainerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.CosmosRepository.ChangeFeed
+{
+    internal interface IContainerChangeFeedProcessor
+    {
+        Task StartAsync();
```
IEvangelist, line 12: I'm curious if a `CancellationToken` should flow through start/stop

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+            _logger.LogInformation("Starting change feed processor for container {ContainerName}", itemContainer.Id);
+
+            await _processor.StartAsync();
+
+            _logger.LogInformation("Successfully started change feed processor for container {ContainerName}",
+                itemContainer.Id);
+        }
+
+        internal async Task OnChanges(IReadOnlyCollection<JObject> changes, CancellationToken cancellationToken,
+            string containerName)
```
IEvangelist, line 70:
```suggestion
        internal async Task OnChangesAsync(
            IReadOnlyCollection<JObject> changes,
            CancellationToken cancellationToken,
            string containerName)
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        public IReadOnlyList<Type> ItemTypes { get; }
+
+        public async Task StartAsync()
+        {
+            Container itemContainer = await _containerService.GetContainerAsync(ItemTypes);
+            Container leaseContainer = await _leaseContainerProvider.GetLeaseContainerAsync();
+
+            _processor = itemContainer.GetChangeFeedProcessorBuilder<JObject>("cosmos-repository-pattern-processor",
+                    (changes, token) => OnChanges(changes, token, itemContainer.Id))
```
IEvangelist, line 55:
```suggestion
                    (changes, token) => OnChangesAsync(changes, token, itemContainer.Id))
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+                if (itemType is null)
+                {
+                    _logger.LogDebug(
+                        "No change feed processor registered for type {ItemType} in container {ContainerName}",
+                        type.Value<string>(), containerName);
+
+                    continue;
+                }
+
+                await InvokeHandler(itemType, change, cancellationToken);
```
IEvangelist, line 94:
```suggestion
                await InvokeHandlerAsync(itemType, change, cancellationToken);
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+
+        private async Task InvokeHandler(Type itemType, JObject instance, CancellationToken cancellationToken)
```
IEvangelist, line 98:
```suggestion
        private async Task InvokeHandlerAsync(
            Type itemType,
            JObject instance,
            CancellationToken cancellationToken)
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            _logger.LogDebug("Invoking IItemChangeFeedProcessor's ({ProcessorsCount}) for item type {ItemType}",
+                handlers.Count(), itemType);
+
+            foreach (object handler in handlers)
+            {
+                try
+                {
+                    object response = handlerType.GetMethod("HandleAsync")?
+                        .Invoke(handler, new[] {item, cancellationToken});
```
IEvangelist, line 122:
```suggestion
                        .Invoke(handler, new[] { item, cancellationToken });
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+                catch (Exception e)
+                {
+                    _logger.LogError(e,
+                        "Failed to handle change for item of type {ItemType} when invoking IItemChangeFeedProcessor {ProcessorTypeName}",
+                        itemType, handler.GetType().Name);
+                }
+            }
+        }
+
+        private Task OnError(Exception exception, string containerName)
```
IEvangelist, line 138:
```suggestion
        private Task OnErrorAsync(Exception exception, string containerName)
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+        public async Task StartAsync()
+        {
+            Container itemContainer = await _containerService.GetContainerAsync(ItemTypes);
+            Container leaseContainer = await _leaseContainerProvider.GetLeaseContainerAsync();
+
+            _processor = itemContainer.GetChangeFeedProcessorBuilder<JObject>("cosmos-repository-pattern-processor",
+                    (changes, token) => OnChanges(changes, token, itemContainer.Id))
+                .WithLeaseContainer(leaseContainer)
+                .WithInstanceName(_changeFeedOptions.InstanceName)
+                .WithErrorNotification((token, exception) => OnError(exception, itemContainer.Id))
```
IEvangelist, line 58:
```suggestion
                .WithErrorNotification((token, exception) => OnErrorAsync(exception, itemContainer.Id))
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+
+        private async Task InvokeHandler(Type itemType, JObject instance, CancellationToken cancellationToken)
+        {
+            object item = instance.ToObject(itemType);
+
+            Type handlerType = null;
+
+            if (Handlers.ContainsKey(itemType) is false)
```
IEvangelist, line 104: Where does `Handlers` come from?

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+            IEnumerable<object> handlers = _serviceProvider.GetServices(handlerType).ToList();
+
+            _logger.LogDebug("Invoking IItemChangeFeedProcessor's ({ProcessorsCount}) for item type {ItemType}",
+                handlers.Count(), itemType);
+
+            foreach (object handler in handlers)
+            {
+                try
+                {
+                    object response = handlerType.GetMethod("HandleAsync")?
```
IEvangelist, line 121:
```suggestion
                    object response = handlerType.GetMethod("HandleAsync")?
```

I'm not super excited by the magic string for the method name here. If it changes, this will break - it's a bit fragile. Since we're down casting to an `object` collection from the service provider of a given type, does that raise any alarms from a DI perspective? Trying to think through this a bit more. What other implications are there? When you step-debug through this, does it do exactly what you imagine?

Microsoft.Azure.CosmosRepository/src/ChangeFeed/InMemory/ChangeFeedItemArgs.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.CosmosRepository.ChangeFeed.InMemory
+{
+    internal class ChangeFeedItemArgs<TItem> where TItem : IItem
+    {
+        public IEnumerable<TItem> ItemChanges { get; set; }
```
IEvangelist, line 10: Does this need to be public `set`?
```suggestion
        public IEnumerable<TItem> ItemChanges { get; private set; }
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/InMemory/ChangeFeedItemArgs.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public ChangeFeedItemArgs(IEnumerable<TItem> itemChanges)
+        {
+            ItemChanges = itemChanges;
+        }
+
+        public ChangeFeedItemArgs(TItem item)
+        {
+            ItemChanges = new[] { item };
+        }
```
IEvangelist, line 20:
```suggestion
        public ChangeFeedItemArgs(IEnumerable<TItem> itemChanges) =>
            ItemChanges = itemChanges;

        public ChangeFeedItemArgs(TItem item) =>
            ItemChanges = new[] { item };
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/Providers/DefaultChangeFeedContainerProcessorProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly IServiceProvider _serviceProvider;
+        private readonly IChangeFeedOptionsProvider _changeFeedOptionsProvider;
+
+        public DefaultChangeFeedContainerProcessorProvider(IOptionsMonitor<RepositoryOptions> optionsMonitor,
+            ICosmosContainerService containerService,
+            ILeaseContainerProvider leaseContainerProvider,
+            ILoggerFactory loggerFactory,
+            IServiceProvider serviceProvider,
+            IChangeFeedOptionsProvider changeFeedOptionsProvider)
+        {
```
IEvangelist, line 31:
```suggestion
        public DefaultChangeFeedContainerProcessorProvider(
            IOptionsMonitor<RepositoryOptions> optionsMonitor,
            ICosmosContainerService containerService,
            ILeaseContainerProvider leaseContainerProvider,
            ILoggerFactory loggerFactory,
            IServiceProvider serviceProvider,
            IChangeFeedOptionsProvider changeFeedOptionsProvider)
        {
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/Providers/DefaultLeaseContainerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+        private const string LeaseContainerName = "lease";
+        private const string LeastContainerPartitionKey = "/id";
+
+        public DefaultLeaseContainerProvider(
+            ICosmosClientProvider cosmosClientProvider,
+            IOptionsMonitor<RepositoryOptions> optionsMonitor)
+        {
+            _repositoryOptions = optionsMonitor.CurrentValue;
+            _cosmosClientProvider = cosmosClientProvider;
+            _lazyContainer = new Lazy<Task<Container>>(BuildLeastContainer);
```
IEvangelist, line 27:
```suggestion
            _lazyContainer = new Lazy<Task<Container>>(BuildLeaseContainer);
```

Microsoft.Azure.CosmosRepository/src/ChangeFeed/Providers/DefaultLeaseContainerProvider.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public Task<Container> GetLeaseContainerAsync() =>
+            _lazyContainer.Value;
+
+        private async Task<Container> BuildLeastContainer()
```
IEvangelist, line 33:
```suggestion
        private async Task<Container> BuildLeaseContainer()
```

IEvangelist (approved):
Approving, but asking lots of questions, proposing newly found suggestions and more styling nits.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/ChangeFeedOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        internal ChangeFeedOptions(Type itemType)
+        {
+            ItemType = itemType;
+        }
+
+        /// <summary>
+        /// The instance name of the processor.
+        /// </summary>
+        public string InstanceName { get; set; } = "default";
```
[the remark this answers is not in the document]
mumby0168: This isn't configurable via the options pattern, this can only be configured as shown below

```
_repositoryOptions.ContainerBuilder.Configure<TestItem>(builder =>
{
    builder.WithContainer("a");
    builder.WithChangeFeedMonitoring(options => options.InstanceName = "abc");
});
```

it's on a per-item basis

Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
```diff
@@ -136,5 +139,22 @@ public ContainerOptionsBuilder WithAutoscaleThroughput(int maxAutoScaleThroughpu
...
             return this;
         }
+
+        /// <summary>
+        /// Adds monitoring of the change feed for the given <see cref="IItem"/>
+        /// </summary>
+        /// <param name="optionsActions">An action to configure the change feed for the given container.</param>
+        /// <returns>Instance of <see cref="ContainerOptionsBuilder"/></returns>
+        /// <remarks>The options configured here are for the container, not just the <see cref="IItem"/> be aware if item's share a container they will share the same change feed options.</remarks>
+        public ContainerOptionsBuilder WithChangeFeedMonitoring(Action<ChangeFeedOptions> optionsActions = null)
```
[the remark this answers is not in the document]
mumby0168: The last call would win? Or do you think if these are already set we should throw?

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultChangeFeedService.cs
```diff
@@ -0,0 +1,40 @@
...
+        public DefaultChangeFeedService(IChangeFeedContainerProcessorProvider changeFeedContainerProcessorProvider)
+        {
+            _processors = new List<IContainerChangeFeedProcessor>();
+            _changeFeedContainerProcessorProvider = changeFeedContainerProcessorProvider;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            _processors = _changeFeedContainerProcessorProvider.GetProcessors();
+            return Task.WhenAll(_processors.Select(x => x.StartAsync()));
```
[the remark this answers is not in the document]
mumby0168: Yes, will update.

Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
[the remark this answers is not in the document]
IEvangelist: I'm fine with whatever approach we've adopted before for similar situations, was more or less curious.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/IContainerChangeFeedProcessor.cs
mumby0168: It hits a bit of a dead-end, when starting the change feed processor I have had to add this but not sure if gives many benefits what do you think?

```csharp
 public async Task StartAsync(CancellationToken cancellationToken)
{
    Container itemContainer = await _containerService.GetContainerAsync(ItemTypes);
    Container leaseContainer = await _leaseContainerProvider.GetLeaseContainerAsync();

    _processor = itemContainer.GetChangeFeedProcessorBuilder<JObject>("cosmos-repository-pattern-processor",
            (changes, token) => OnChanges(changes, token, itemContainer.Id))
        .WithLeaseContainer(leaseContainer)
        .WithInstanceName(_changeFeedOptions.InstanceName)
        .WithErrorNotification((token, exception) => OnError(exception, itemContainer.Id))
        .Build();

    cancellationToken.Register(() => StopAsync(cancellationToken));

    _logger.LogInformation("Starting change feed processor for container {ContainerName}", itemContainer.Id);

    await _processor.StartAsync();

    _logger.LogInformation("Successfully started change feed processor for container {ContainerName}",
        itemContainer.Id);
}
```

Microsoft.Azure.CosmosRepository/src/Builders/ContainerOptionsBuilder.cs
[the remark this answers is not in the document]
mumby0168: This like the other takes the rule of last call to the method wins, it would be odd to see this called more than once I think.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
mumby0168: It's just a static collection of constructed handler types, it means you only construct the generic type once.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
mumby0168: Yes I was not over the moon about this either, but if I am honest I think it's a trade-off we may have to take, there was no way (at least that I could see) when having one change feed processor per container, that could handle multiple `IItem`s to flow down a generic `TItem` property to neatly resolve this from DI by calling something like `_provider.GetRequiredService<IItemChangeFeedProcessor<TItem>>();` which was my initial intention, this was the best I could do, as far as debugging it behaves the same way you would see the normal call as shown but I agree the floating string feels a bit meh, however, we control the interface so I think that makes it a lot safer. Thoughts?

Microsoft.Azure.CosmosRepository/src/ChangeFeed/IContainerChangeFeedProcessor.cs
IEvangelist: I'm not certain either. I'd just hate to have a bug reported at a later time where someone couldn't get cancellation to work as expected.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
IEvangelist: That's fair, I was just pushing on it a bit as I was curious what alternatives have been looked at - and how you felt about it. Thanks

Microsoft.Azure.CosmosRepository/src/ChangeFeed/IContainerChangeFeedProcessor.cs
mumby0168: Yes I agree, I will put a sample together to show the cancellation as well as testing it, will make sure this happens before a release.

Microsoft.Azure.CosmosRepository/src/ChangeFeed/DefaultConatinerChangeFeedProcessor.cs
mumby0168: Okay cool, I agree with you tho, doesn't look super clean but not sure it does much harm by the same token.

IEvangelist (approved):

IEvangelist (approved):

Microsoft.Azure.CosmosRepository/src/InMemoryRepository.cs
IEvangelist: An `Action<T>` is also a multicast delegate, it just doesn't add the `object` parameter to the method signature.

mumby0168: Ah, that is cool to know, learn something new everyday!

Krumelur:
@mumby0168 I just updated to latest to see if the problem with the `Type` property is fixed, but now I noticed that with #163  you locked down `ICosmosContainerService`. It went from `public` to `internal`. I'm using this interface to execute an arbitrary query against my container. Now, this is no longer possible. What's the strategy behind locking everything down so much?

mumby0168:
> @mumby0168 I just updated to latest to see if the problem with the `Type` property is fixed, but now I noticed that with #163 you locked down `ICosmosContainerService`. It went from `public` to `internal`. I'm using this interface to execute an arbitrary query against my container. Now, this is no longer possible. What's the strategy behind locking everything down so much?

This was exposed by mistake. However, I think we could re-expose this. What was it you were using it for out of curiosity we to do try and avoid breaking changes Apologies.

@IEvangelist any objections to re-exposing this interface?

IEvangelist:
> > @mumby0168 I just updated to latest to see if the problem with the `Type` property is fixed, but now I noticed that with #163 you locked down `ICosmosContainerService`. It went from `public` to `internal`. I'm using this interface to execute an arbitrary query against my container. Now, this is no longer possible. What's the strategy behind locking everything down so much?
> 
> This was exposed by mistake. However, I think we could re-expose this. What was it you were using it for out of curiosity we to do try and avoid breaking changes Apologies.
> 
> @IEvangelist any objections to re-exposing this interface?

I'm really curious what arbitrary queries you were running directly against the container? What feature gaps do we have? What bits of information are hard to get at? I'd rather create helper functions to expose the specific bits you're trying to get at, than re-expose something that was intended to be `internal`.

> What's the strategy behind locking everything down so much?

Ideally, when we design something, we'll expose the least amount of API surface such that the consumer isn't bothered by complexity, or clutter. If everything for example was public, it would be REALLY hard to figure out what you should use. We intentionally, and mindfully make certain implementation details `internal` as to not expose them. I think we can compromise a bit, and thinking ahead, we could probably expose some of the native `CosmosClient` entirely disconnected from our wrapper just through DI. I'm a bit torn though. @mumby0168 what are your thoughts?

mumby0168:
yes I am cautious as we don't want to have breaking change so having the minimal amount of functionality exposed to the consumer makes us more capable of making changes underneath. I would like to know what extra functions you would need to see if we can provide them. 

That said, the cosmos container service in this example can be useful to plug gaps - such as the transactional batch bits and there are a few more that direct access to a `Container` would make easier for the time being. I feel as since this was `public` albeit a mistake since it was we maybe should change it back since people are now depending on it.

It is a low risk this one I think.

@IEvangelist @Krumelur

IEvangelist:
@mumby0168, I'm good with that.

mumby0168:
> @mumby0168, I'm good with that.

Okay I'll raise a PR shortly to reexpose that interface then. I think your right tho we should start with as much internal as possible and as and when allow access to interfaces where we see fit 👍🏼

Issue #164, opened 2022-01-09
Functional/Integration Tests


mumby0168:
It would be good if we could look at adding some tests that run against the real cosmos DB service to assert that the functions we provide are working between PR's and commits to avoid releasing issues in the library.

I see a few options for this:

* Run the tests against a provisioned cosmos account in the cloud, we could utilize the free tier of 1000 RUs (Connection String as an environment variables in a GA)
* Run these tests against the simulator hosted via docker in a GitHub action

I think either option would have to delete the database after any tests and re-create it next. These could tie the library together using Microsoft's `IServiceProvider`.

Once this is set up we can write some tests that act like integration/functional tests of the library to confirm critical flows always work when the library is modified.

ylr-research:
I like option 1

ylr-research:
Never mind. Answer is above

ylr-research:
@mumby0168 any specific account you prefer to use for testing?

mumby0168:
> @mumby0168 any specific account you prefer to use for testing?

No, we'd have to chat to @IEvangelist around this I think, we'd not want to run these over & over depending on there RU consumption we would want to avoid this costing us if possible I think but they would provide a large benefit I think, but like with any functional tests we want this to be solid and repeatable and certainly not flakey which is something we would have to consider from the get go. Sure it's possible tho, going to have a look into this soon 😃

dcuccia:
Can we ensure for a class of functional tests that only one instance runs concurrently? If not, we could run them against the real DB as a scheduled cron job (nightly or hourly), so that we're ensured to not go over the limit.

mumby0168:
> Can we ensure for a class of functional tests that only one instance runs concurrently? If not, we could run them against the real DB as a scheduled cron job (nightly or hourly), so that we're ensured to not go over the limit.

We could look to do that if they get that big I think to start with we should be okay, they only need to be smoke tests really just to make sure the core bits of the library function properly, so should be okay to run on a PR I think maybe they have to manually run. We don't get that many PRs at the same time.

mumby0168:
The basics of these are now in place, I want to add some contribution documentation to this. See the [Accepetance Tests here](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/tree/main/tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests)

mumby0168:
These are running as part of the CI build on main now. Going to close this issue.

PR #169, opened 2022-01-11
Basic Acceptance Testing

mumby0168:
Related #164 

This is the general idea of what I see the acceptance tests to look like, thoughts @IEvangelist?

There are a few points for discussion below:

1. How often do we run these? (nightly, for every PR? for every merge to master?), I feel there are trade off's with each approach
2. Where do we set up the cosmos instance to test against? (The free tier of 1000 RUs should keep this FREE.
3. We need to find a way to exclude these tests being run for every case, this should be opt-in only I think.

I have added an example test that gives an idea that these tests are not just to test a single `CreateAsync` for example but to test the library doing realistic sets of operations and working as expected.


tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/CosmosRepositoryAcceptanceTest.cs
```diff
@@ -0,0 +1,69 @@
...
+public abstract class CosmosRepositoryAcceptanceTest
+{
+    private const string ProductsInfoContainer = "products-info";
+    private const string DefaultPartitionKey = "/partitionKey";
+    protected const string TechnologyCategoryId = "Techonology";
+
+    protected readonly IRepository<Product> _productsRepository;
+    protected readonly IRepository<Rating> _ratingsRepository;
+
+    protected EquivalencyAssertionOptions<Product> DefaultProductEquivalencyOptions(EquivalencyAssertionOptions<Product> options)
```
IEvangelist, line 23:
```suggestion
    protected EquivalencyAssertionOptions<Product> DefaultProductEquivalencyOptions(
        EquivalencyAssertionOptions<Product> options)
```

tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/CosmosRepositoryAcceptanceTest.cs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+
+    protected CosmosRepositoryAcceptanceTest(Action<RepositoryOptions> builderOptions)
+    {
+        ConfigurationBuilder config = new();
+        config.AddUserSecrets<CosmosRepositoryAcceptanceTest>();
+
+        ServiceCollection services = new();
+
```
IEvangelist, line 40:
```suggestion
```

tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/CosmosRepositoryAcceptanceTest.cs
```diff
@@ -0,0 +1,69 @@
...
+    protected EquivalencyAssertionOptions<Product> DefaultProductEquivalencyOptions(EquivalencyAssertionOptions<Product> options)
+    {
+        options.Excluding(x => x.Etag);
+        options.Excluding(x => x.CreatedTimeUtc);
+        options.Excluding(x => x.LastUpdatedTimeRaw);
+        options.Excluding(x => x.LastUpdatedTimeUtc);
+
+        return options;
+    }
+
```
IEvangelist, line 32:
```suggestion
```

IEvangelist (approved):
I like it, clean and simple.

IEvangelist:
> How often do we run these? (nightly, for every PR? for every merge to master?), I feel there are trade off's with each approach

As for the number of times they run, we could start by running them `on` every `push` to `main`. And if that's too much, we can reassess and tune it back as needed.

> We need to find a way to exclude these tests being run for every case, this should be opt-in only I think.

We use xUnit, we should be able to use categories.

> Where do we set up the cosmos instance to test against? (The free tier of 1000 RUs should keep this FREE.

I'll reach out to the Cosmos team and see if they might be able to help us out with a test account for this.

mumby0168:
> > How often do we run these? (nightly, for every PR? for every merge to master?), I feel there are trade off's with each approach
> 
> As for the number of times they run, we could start by running them `on` every `push` to `main`. And if that's too much, we can reassess and tune it back as needed.
> 
> > We need to find a way to exclude these tests being run for every case, this should be opt-in only I think.
> 
> We use xUnit, we should be able to use categories.
> 
> > Where do we set up the cosmos instance to test against? (The free tier of 1000 RUs should keep this FREE.
> 
> I'll reach out to the Cosmos team and see if they might be able to help us out with a test account for this.

Okay that sound's great, let me know what the outcome is 😃

I'll work on adding the xUnit categories to this.

I agree let's run it on master only for now and then we can add a manual run for this if we think a PR branch should be checked before a merge.

IEvangelist (dismissed):

IEvangelist (approved):

Issue #165, opened 2022-01-10
Allow for sorting when doing paginated requests


filipmhpersson:
**Is your feature request related to a problem? Please describe.**

We need to be able to sort paginated queries. I see this as a critical feature needed for paging. Lists usually needs to be sorted and when the requests is paginated that the needs to be done serverside otherwise we always need to get all of the data, sort it and then page it, instead of letting the cosmos db serve the paginated response.

I'm not sure if I have missed something in the interface but apart from direct SQL queries I don't see a possibility to do sorting when using PageAsync. And if I understood correctly doing direct queries will ruin the possibility to use the in memory database for testing. Which is a really nice feature to have.

**Describe the solution you'd like**
The biggest question for this feature is how update the IRepository interface to include sorting. 

My suggestion would be to look at the specification pattern that can be seen here:

https://github.com/ardalis/Specification

But limit it to whats needed here, since the ardalis specification implementation is sql + EF specific. This would also allow for a scalable way to implement more features in the future. This is what I think would be needed for a specification pattern that's not too complex to implement and would give value right now:

 ```  c#
 //Setup classes
 public interface ISpecification<T>
    {
        IEnumerable<Expression<Func<T, bool>> WhereExpressions { get; }
        IEnumerable<Expression<Func<T, object?>>, OrderTypeEnum> OrderExpressions { get; }

        int? Take { get; }
        int? Skip { get; }

        string? ContinuationToken{ get; }
    }

  public enum OrderTypeEnum
    {
        OrderBy = 1,
        OrderByDescending = 2,
        ThenBy = 3,
        ThenByDescending = 4
    }
    
//Page method:

public async ValueTask<IPageQueryResult<TItem>> PageAsync(
            ISpecification<TItem>,
            CancellationToken cancellationToken = default)
 ```

**Describe alternatives you've considered**
It is also possible to extend the page method to include the order by clause:
`  IEnumerable<Expression<Func<T, object?>> ` 
But I don't think this is a scaleable way to keep developing the repository. It will probably lead to a method that just keeps growing in size. 

Another solution is to got the IQueryable path but I agree with the conclussions reached here: https://github.com/IEvangelist/azure-cosmos-dotnet-repository/issues/88 that opening the door to IQueryable will add lots of complexity for the small subset of features thats needed. 

**Additional context**
If there currently is another way to sort results for a paged query please let me know, I have tried to look everyone but then I missed it.

mumby0168:
[...]

PR #171, opened 2022-01-13
Users/filipmhpersson/sortspecification

filipmhpersson:
Implementation of specification pattern
Based on discussion in https://github.com/IEvangelist/azure-cosmos-dotnet-repository/issues/165.

I'm not sure how we will get the specification working together with continuation token and skip/take and have a return type that is the same for both ways of paging. I don't see a way for how we can specify CurrentPage when using continuation token, but maybe there is a way?

The specific implementation with evaluators is very inspired by this repository https://github.com/ardalis/Specification, but we cannot use that out of the box since it's very SQL specific and this is for cosmos, so lots of changes and been made and it will keep changing. But I think it should be given credit to in some way if we decide to add this pattern to the repository.

Still know WIP:
XML Comments
Make lots of public classes internal
Specific tests for specification function


Microsoft.Azure.CosmosRepository.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
```
filipmhpersson, line 1: Sorry about this file is it because I'm using Visual Studio 2022?

I'll make sure to clean it up if it's a problem

src/Microsoft.Azure.CosmosRepository/Specification/BaseSpecification.cs
```diff
@@ -0,0 +1,51 @@
...
+        public BaseSpecification(Expression<Func<T, bool>>[] filters)
+        {
+            Query = new SpecificationBuilder<T>(this);
+            foreach(Expression<Func<T, bool>> filter in filters)
+            {
+                Query.Where(filter);
+            }
+        }
+        /// <inheritdoc/>
+        public IEnumerable<WhereExpressionInfo<T>> WhereExpressions { get; } = new List<WhereExpressionInfo<T>>();
```
mumby0168, line 39: Would these want to be `IReadOnlyList<T>` ? Do we want people to edit these collections directly or use the builder?

src/Microsoft.Azure.CosmosRepository/IRepository.cs
```diff
@@ -266,6 +267,18 @@ public interface IRepository<TItem> where TItem : IItem
             string continuationToken = null,
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Offers a load more paging implementation for infinite scroll scenarios.
```
mumby0168, line 271: I would maybe explain the `ISpecification<T>` here ?

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
```diff
@@ -346,6 +349,29 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
 
+        public async ValueTask<IPageQueryResult<TItem>> PageAsync(ISpecification<TItem> specification, CancellationToken cancellationToken = default)
+        {
+            await Task.CompletedTask;
+
+            IQueryable<TItem> query = Items.Values.Select(DeserializeItem).AsQueryable()
+                .Where(item => item.Type == typeof(TItem).Name);
+
+            int pageSize = specification.PageSize;
+            ISpecificationEvaluator specificationEvaluator = new SpecificationEvaluator();
```
mumby0168, line 360: Can this be injected via DI?

mumby0168 (commented):
I love how this is looking, a really huge feature, a few little bits a few more examples would certainly help reinforce how this could work.

I am keen to explore possibly moving some of those default `Specification<T>` implementations you have done in the sample app into the repo as possible base classes just for ease of use for a consumer.

mumby0168:
> Implementation of specification pattern Based on discussion in #165.
> 
> I'm not sure how we will get the specification working together with continuation token and skip/take and have a return type that is the same for both ways of paging. I don't see a way for how we can specify CurrentPage when using continuation token, but maybe there is a way?
> 
> The specific implementation with evaluators is very inspired by this repository https://github.com/ardalis/Specification, but we cannot use that out of the box since it's very SQL specific and this is for cosmos, so lots of changes and been made and it will keep changing. But I think it should be given credit to in some way if we decide to add this pattern to the repository.
> 
> Still know WIP: XML Comments Make lots of public classes internal Specific tests for specification function

In terms of the continuation token's and page number this is not something that could be supported. The token is the page marker in affect, does that make sense?

samples/Paging/Program.cs
```diff
@@ -7,12 +7,13 @@
...
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Paging.Specifications;
 
 ConfigurationBuilder configuration = new();
 
 ServiceProvider provider = new ServiceCollection().AddCosmosRepository(options =>
     {
-        options.CosmosConnectionString = Environment.GetEnvironmentVariable("CosmosConnectionString");
+        options.CosmosConnectionString = "AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";
```
IEvangelist, line 16: Wow, probably not an ideal place to have a hard coded string that contains sub key. If you haven't already, you should remove this immediately - and then go and delete your corresponding resource (or at the very least regenerate the keys).

filipmhpersson: Oh will cleanup. but it's no big deal from security aspect since this is the default and only possible accountkey for the cosmos db emulator.
https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator?tabs=ssl-netstd21#authenticate-requests

filipmhpersson:
> > Implementation of specification pattern Based on discussion in #165.
> > I'm not sure how we will get the specification working together with continuation token and skip/take and have a return type that is the same for both ways of paging. I don't see a way for how we can specify CurrentPage when using continuation token, but maybe there is a way?
> > The specific implementation with evaluators is very inspired by this repository https://github.com/ardalis/Specification, but we cannot use that out of the box since it's very SQL specific and this is for cosmos, so lots of changes and been made and it will keep changing. But I think it should be given credit to in some way if we decide to add this pattern to the repository.
> > Still know WIP: XML Comments Make lots of public classes internal Specific tests for specification function
> 
> In terms of the continuation token's and page number this is not something that could be supported. The token is the page marker in affect, does that make sense?

Yes  but I think I have an idea based on your other comments.

Maybe we can look at making it 
ISpecification<TItem,TResult>
Then you can have
BaseSpecification<TItem,IEnumerable<T>>
ConitutationTokenSpecification<IItem,Page<T>>
SkipTakeSpecification<Item,PageQuery<T>>

And remove paging functionality from BaseSpecification and implement specific versions of it for the continutationtoken and skiptake. Think this would be a good way to communicate paging functionality to developers using specifications?

mumby0168:
> > > Implementation of specification pattern Based on discussion in #165.
> 
> > > I'm not sure how we will get the specification working together with continuation token and skip/take and have a return type that is the same for both ways of paging. I don't see a way for how we can specify CurrentPage when using continuation token, but maybe there is a way?
> 
> > > The specific implementation with evaluators is very inspired by this repository https://github.com/ardalis/Specification, but we cannot use that out of the box since it's very SQL specific and this is for cosmos, so lots of changes and been made and it will keep changing. But I think it should be given credit to in some way if we decide to add this pattern to the repository.
> 
> > > Still know WIP: XML Comments Make lots of public classes internal Specific tests for specification function
> 
> > 
> 
> > In terms of the continuation token's and page number this is not something that could be supported. The token is the page marker in affect, does that make sense?
> 
> 
> 
> Yes  but I think I have an idea based on your other comments.
> 
> 
> 
> Maybe we can look at making it 
> 
> ISpecification<TItem,TResult>
> 
> Then you can have
> 
> BaseSpecification<TItem,IEnumerable<T>>
> 
> ConitutationTokenSpecification<IItem,Page<T>>
> 
> SkipTakeSpecification<Item,PageQuery<T>>
> 
> 
> 
> And remove paging functionality from BaseSpecification and implement specific versions of it for the continutationtoken and skiptake. Think this would be a good way to communicate paging functionality to developers using specifications?
> 
> 

Yes I think that sounds ace, then you are right it is really clear what you are capable of doing and how it intends to work.

samples/Paging/Program.cs
mumby0168: For the future you can set env variables for a test runner, I use rider and you can do it there but fairly sure you can do the same for VS 2022 with a run settings file or something like that, @IEvangelist may know more 😃

ylr-research:
@mumby0168 if I'm not mistaken offset and cursor pattern are mutually exclusive. They serve different purposes. Have pros and cons. Offset is for smaller sets and allow pagination through skip/take, whereas cursor is for large sets and uses the continuation token as a pointer. Offset you can say go to page 3 while with cursor you can't.

mumby0168:
> @mumby0168 if I'm not mistaken offset and cursor pattern are mutually exclusive. They serve different purposes. Have pros and cons. Offset is for smaller sets and allow pagination through skip/take, whereas cursor is for large sets and uses the continuation token as a pointer. Offset you can say go to page 3 while with cursor you can't.

Exactly, we'll put.

IEvangelist:
@all-contributors please add @filipmhpersson for code and tests

filipmhpersson:
@mumby0168 I have updated the PR now with a more generic management of return types. This allows for different specification to specify the return type, so it's either a list, a IPage or IPageQueryResult. In general this feels much better for a clear seperation between offset/limit and continuation token patterns. I still need to break out UseContinutationToken and PageNumber from the SpecificationBuilder and make sure they are only useable from
ContinuationTokenSpecification and PageNumberPageSizeSpecification (this one really needs a new name :)). Also some cleanup from generics in the latest PR is required. I just wanted to prove the pattern before cleaning it up. But there is no real reason for the OrderedEvaluator for example to have a generic return type.

ylr-research:
You can name it OffsetSpecification

mumby0168:
> @mumby0168 I have updated the PR now with a more generic management of return types. This allows for different specification to specify the return type, so it's either a list, a IPage or IPageQueryResult. In general this feels much better for a clear seperation between offset/limit and continuation token patterns. I still need to break out UseContinutationToken and PageNumber from the SpecificationBuilder and make sure they are only useable from
> ContinuationTokenSpecification and PageNumberPageSizeSpecification (this one really needs a new name :)). Also some cleanup from generics in the latest PR is required. I just wanted to prove the pattern before cleaning it up. But there is no real reason for the OrderedEvaluator for example to have a generic return type.

I will give this a review tomorrow, thanks again for your contribution.

src/Microsoft.Azure.CosmosRepository/IQueryResult.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.CosmosRepository
+{
+    /// <summary>
+    /// Represent a full set of data from a cosmos query
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public interface IQueryResult<out T> where T : IItem
```
mumby0168, line 14: We tend to use the generic argument name `TItem` on all our other geneirc implementations

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
```diff
@@ -346,6 +349,23 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
+            query = specificationEvaluator.GetQuery(query, specification);
+
+            int countResponse =query.Count();
```
mumby0168, line 364:
```diff
+ int countResponse = query.Count();
- int countResponse =query.Count();
```

samples/Paging/Specifications/ContinuationTokenSpecification.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.CosmosRepository.Specification;
+
+namespace Paging.Specifications
+{
+    internal class ContinuationTokenSpecificationImplementation : ContinuationTokenSpecification<Person>
+    {
+        public ContinuationTokenSpecificationImplementation(string continutationToken, int pageSize):
```
mumby0168, line 15:
```suggestion
        public ContinuationTokenSpecificationImplementation(string continuationToken, int pageSize):
```

samples/Paging/Program.cs
```diff
@@ -79,6 +108,32 @@ async Task BasicScrollingAsync()
     Console.WriteLine($"Total Charge {totalCharge} RU's");
 }
 
+
+async Task BasicSpecificationScrollingAsync()
+{
+    double totalCharge = 0;
+
+    ContinuationTokenSpecificationImplementation specification = new(null,pageSize: 25);
```
mumby0168, line 116: It would be good to see examples using the new base types implemented inside the library.

samples/Paging/Specifications/ContinuationTokenSpecification.cs
mumby0168: In fact, can we delete this now we have the implementation of this inside the library?

samples/Paging/Specifications/PageNumberPageSizeSpecification.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.CosmosRepository.Specification;
+
+namespace Paging.Specifications
+{
+    internal class PageNumberPageSizeSpecificationImplementation : PageNumberPageSizeSpecification<Person>
```
mumby0168, line 13: Can this be deleted now?

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -391,6 +399,31 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
                 result.Charge + countResponse.RequestCharge);
         }
 
+        public async ValueTask<TResult> GetAsync<TResult>(ISpecification<TItem, TResult> specification, CancellationToken cancellationToken = default)
+            where TResult : IQueryResult<TItem>
+        {
+            Container container = await _containerProvider.GetContainerAsync().ConfigureAwait(false);
+
```
mumby0168, line 406: nit: extra spacing here not needed.

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -391,6 +399,31 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
+
+            QueryRequestOptions options = new() { };
```
mumby0168, line 408: Can we remove this if we no query options?

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -391,6 +399,31 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
+            if (specification.UseContinutationToken)
+            {
+                options.MaxItemCount = specification.PageSize;
+            }
+
+            IQueryable<TItem> query =
+                    container.GetItemLinqQueryable<TItem>(requestOptions: options,continuationToken: specification.ContinutationToken)
```
mumby0168, line 415:
```suggestion
                    container.GetItemLinqQueryable<TItem>(continuationToken: specification.ContinutationToken)
```

src/Microsoft.Azure.CosmosRepository/QueryResult.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.CosmosRepository
```
mumby0168, line 8: Need to update the XML comments here.

src/Microsoft.Azure.CosmosRepository/Specification/Builder/ISpecificationBuilder.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text;
+
+namespace Microsoft.Azure.CosmosRepository.Specification
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <typeparam name="TResult"></typeparam>
+    public interface ISpecificationBuilder<T, TResult>
```
mumby0168, line 15: Does this interface need to be public? If we are exposing the bae class?

src/Microsoft.Azure.CosmosRepository/Specification/Builder/OrderedSpecificationBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <param name="specification"></param>
+        public OrderedSpecificationBuilder(BaseSpecification<T, TResult> specification)
+        {
+            Specification = specification;
+        }
+    }
+    /// <summary>
+    /// 
+    /// </summary>
+    public static class OrderedBuilderExtensions
```
mumby0168, line 30: I would pop these in an `Extensions` folder under the `Specifications` folder

src/Microsoft.Azure.CosmosRepository/Specification/Builder/SpecificationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /// <summary>
+        /// Specify the query result will be ordered by <paramref name="orderExpression"/> in a descending order
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="specificationBuilder"></param>
+        /// <param name="orderExpression"></param>
+        public static IOrderedSpecificationBuilder<T, TResult> OrderByDescending<T, TResult>(
```
mumby0168, line 74: This seems to imply you could do something such as `Query.Orderby(x => ...)` should you be able to do this before a `.Where` clause?

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/PagingEvaluator.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Azure.CosmosRepository.Specification.Evaluator
```
mumby0168, line 9: XML comments

src/Microsoft.Azure.CosmosRepository/Specification/Specificiations/BaseSpecification.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// 
+        /// </summary>
+        protected virtual ISpecificationBuilder<T, TResult> Query { get; }
+        /// <summary>
+        /// Initialize specificaiton and add filters later
+        /// </summary>
+        public BaseSpecification()
+        {
+            Query = new SpecificationBuilder<T, TResult>(this);
```
mumby0168, line 24:
```suggestion
public BaseSpecification() =>
             Query = new SpecificationBuilder<T, TResult>(this);
```

src/Microsoft.Azure.CosmosRepository/Specification/Builder/OrderedSpecificationBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <inheritdoc/>
+        public BaseSpecification<T,TResult> Specification { get; }
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="specification"></param>
+        public OrderedSpecificationBuilder(BaseSpecification<T, TResult> specification)
+        {
+            Specification = specification;
+        }
```
mumby0168, line 25:
```suggestion
        public OrderedSpecificationBuilder(BaseSpecification<T, TResult> specification) => 
            Specification = specification;
        
```

mumby0168 (commented):
Looks good, there are just a few bits I wanted to highlight.

* There are still a few unresolved comments from the last PR review
* It would be good to see some examples of this in use under the `samples` directory it really helps to see the intent of this PR
* There are a few code formatting nits if we have a method one a single line of code we always prefer an expression body. We also aim to not have double sets of whitespace if possible.
* We need to look at updating all the XML comments as well.
* It would be nice to see some acceptance tests written for these new features.


This is a really cool PR and I am looking forward to merging it.

src/Microsoft.Azure.CosmosRepository/Specification/BaseSpecification.cs
filipmhpersson: I think all work should be done via the builders

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
filipmhpersson: Is there a pattern for DI use in the InMemoryRepository? 
It's already using DI in defaultrepository

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/IEvaluator.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface IEvaluator
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        bool IsCriteriaEvaluator { get; }
```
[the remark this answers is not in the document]
filipmhpersson: If the evaluator evaluates the where criteria. I updated the name to IsFilterEvaluator.

The reason this is needed is that we for a Count operation need to only get the CriteriaEvaluator and not the paging/order evaluators.

mumby0168:
@filipmhpersson great work! Love the samples they look really cool, I think they're is just some XML comments to update and maybe some tests to add, and maybe a bit of formatting here and there but on the whole looking great.

Microsoft.Azure.CosmosRepository.sln
IEvangelist: No worries, I already upgraded this to 2022

samples/Specification/Program.cs
```diff
@@ -0,0 +1,113 @@
...
+FullSpecificationSamples fullSpecificationSamples = new(repository);
+
+Console.WriteLine("Continuation Token with query sample");
+await fullSpecificationSamples.FullContinuationTokenSpecificationAsync(10);
+
+Console.WriteLine("Offset by page number query sample");
+await fullSpecificationSamples.FullPageNumberSpecificationAsync(10);
+
+
+
```
IEvangelist, line 55:
```suggestion
```

samples/Specification/Program.cs
```diff
@@ -0,0 +1,113 @@
...
+        Console.WriteLine(person);
+    }
+
+    totalCharge += page.Charge;
+
+    Console.WriteLine($"Last 50 results cost {page.Charge}");
+    Console.WriteLine($"Total Charge {totalCharge} RU's");
+}
+
+
```
IEvangelist, line 95:
```suggestion
```

samples/Specification/Specification.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <ItemGroup>
+        <PackageReference Include="Bogus" Version="33.0.2" />
+        <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
+        <PackageReference Include="System.Linq.Async" Version="5.0.0" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="..\..\src\Microsoft.Azure.CosmosRepository\Microsoft.Azure.CosmosRepository.csproj" />
+    </ItemGroup>
+
```
IEvangelist, line 19:
```suggestion
```

samples/Specification/SpecificationOrderSamples.cs
```diff
@@ -0,0 +1,75 @@
...
+public class SpecificationOrderSamples
+{
+    private readonly IRepository<Person> _repository;
+
+    public SpecificationOrderSamples(IRepository<Person> repository)
+    {
+        _repository = repository;
+    }
+    public async Task BasicOrderAsync()
+    {
```
IEvangelist, line 24:
```suggestion

    public async Task BasicOrderAsync()
    {
```

samples/Specification/SpecificationPagingSamples.cs
```diff
@@ -0,0 +1,67 @@
...
+public class SpecificationPagingSamples
+{
+    private readonly IRepository<Person> _repository;
+
+    public SpecificationPagingSamples(IRepository<Person> repository)
+    {
+        _repository = repository;
+    }
+    public async Task BasicPageAsync()
+    {
```
IEvangelist, line 24:
```suggestion

    public async Task BasicPageAsync()
    {
```

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
IEvangelist: The `InMemoryRepository` technically supports full DI, as it's added to the service collection/provider. You should be able to add it as a .ctor parameter, and DI would provide it.

src/Microsoft.Azure.CosmosRepository/Specification/Builder/OrderedSpecificationBuilder.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="orderBuilder"></param>
+        /// <param name="orderExpression"></param>
+        /// <returns></returns>
+        public static IOrderedSpecificationBuilder<T, TResult> ThenBy<T, TResult>(
+            this IOrderedSpecificationBuilder<T, TResult> orderBuilder,
+            Expression<Func<T, object>> orderExpression)
+            where T: IItem
+            where TResult: IQueryResult<T>
+        {
+            ((List<OrderExpressionInfo<T>>)orderBuilder.Specification.OrderExpressions)
```
IEvangelist, line 48: I'm not sure I like the explicit casting here, if we know the type is for sure this - why do we not expose it as such?

src/Microsoft.Azure.CosmosRepository/Specification/Builder/SpecificationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+namespace Microsoft.Azure.CosmosRepository.Specification
+{
+    class SpecificationBuilder<T,TResult> : ISpecificationBuilder<T,TResult>
+        where T : IItem
+        where TResult : IQueryResult<T>
+
+    {
+        public BaseSpecification<T,TResult> Specification { get; }
+        public SpecificationBuilder(BaseSpecification<T, TResult> specification)
+        {
```
IEvangelist, line 19:
```suggestion
        public BaseSpecification<T, TResult> Specification { get; }
        
        public SpecificationBuilder(BaseSpecification<T, TResult> specification)
        {
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/IEvaluator.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="query"></param>
+        /// <param name="specification"></param>
+        /// <returns></returns>
+        IQueryable<T> GetQuery<T, TResult>(IQueryable<T> query, ISpecification<T, TResult> specification)
+            where T : IItem
+            where TResult : IQueryResult<T>;
+
```
IEvangelist, line 31:
```suggestion
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/OrderEvaluator.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Azure.CosmosRepository.Specification.Evaluator
+{
+    class OrderEvaluator : IEvaluator
+    {
+        private OrderEvaluator() { }
+        public static OrderEvaluator Instance { get; } = new OrderEvaluator();
```
IEvangelist, line 14:
```suggestion

        public static OrderEvaluator Instance { get; } = new();
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/OrderEvaluator.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public bool IsFilterEvaluator { get; } = false;
+
+        public IQueryable<T> GetQuery<T, TResult>(IQueryable<T> query, ISpecification<T, TResult> specification)
+            where T : IItem
+            where TResult : IQueryResult<T>
+        {
+            if(specification.ContinutationToken != null && specification.ContinutationToken != "")
```
IEvangelist, line 22:
```suggestion
            if (specification.ContinutationToken is { Length: > 0 })
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/OrderEvaluator.cs
```diff
@@ -0,0 +1,59 @@
...
+                        OrderTypeEnum.ThenByDescending => orderedQuery.ThenByDescending(orderExpression.KeySelector),
+                        _ => throw new NotImplementedException("Unknown value of OrderTypeEnum")
+                    };
+                }
+
+                if (orderedQuery != null)
+                {
+                    query = orderedQuery;
+                }
+            }
```
IEvangelist, line 53:
```suggestion
            if ((specification.OrderExpressions?.Count(
                x => x.OrderType == OrderTypeEnum.OrderBy ||
                x.OrderType == OrderTypeEnum.OrderByDescending) ?? 0) > 1)
            {
                throw new ArgumentException("Multiple OrderBy expressions found only use one and then chain with ThenBy and TheByDescending");
            }

            IOrderedQueryable<T> orderedQuery = null;
            foreach (OrderExpressionInfo<T> orderExpression in specification.OrderExpressions)
            {
                orderedQuery = orderExpression.OrderType switch
                {
                    OrderTypeEnum.OrderBy => query.OrderBy(orderExpression.KeySelector),
                    OrderTypeEnum.OrderByDescending => query.OrderByDescending(orderExpression.KeySelector),
                    OrderTypeEnum.ThenBy => orderedQuery.ThenBy(orderExpression.KeySelector),
                    OrderTypeEnum.ThenByDescending => orderedQuery.ThenByDescending(orderExpression.KeySelector),
                    _ => throw new NotImplementedException("Unknown value of OrderTypeEnum")
                };
            }

            if (orderedQuery is not null)
            {
                query = orderedQuery;
            }
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/PagingEvaluator.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// 
+    /// </summary>
+    public class PagingEvaluator : IEvaluator
+    {
+        private PagingEvaluator() { }
+        /// <summary>
+        /// 
+        /// </summary>
+        public static PagingEvaluator Instance { get; } = new PagingEvaluator();
```
IEvangelist, line 20:
```suggestion
        public static PagingEvaluator Instance { get; } = new();
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/PagingEvaluator.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <typeparam name="T"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="query"></param>
+        /// <param name="specification"></param>
+        /// <returns></returns>
+        public IQueryable<T> GetQuery<T, TResult>(IQueryable<T> query, ISpecification<T,TResult> specification)
+            where T : IItem
+            where TResult : IQueryResult<T>
+        {
+
```
IEvangelist, line 37:
```suggestion
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/PagingEvaluator.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (specification.UseContinutationToken)
+            {
+                return query;
+            }          
+
+            if(specification.PageNumber.HasValue && specification.PageNumber != 0)
```
IEvangelist, line 43:
```suggestion
            if (specification.PageNumber.GetValueOrDefault() > 0)
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/SpecificationEvaluator.cs
```diff
@@ -0,0 +1,48 @@
...
+           
+        }
+
+        public TResult GetResult<T, TResult>(IReadOnlyList<T> res, ISpecification<T, TResult> specification, int totalCount, double charge, string continuationToken)
+            where T : IItem
+            where TResult : IQueryResult<T>
+        {
+            return specification.PostProcessingAction(res, totalCount, charge, continuationToken);
+        }
+
```
IEvangelist, line 46:
```suggestion
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/WhereEvaluator.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// 
+    /// </summary>
+    public class WhereEvaluator : IEvaluator
+    {
+        private WhereEvaluator() { }
+        /// <summary>
+        /// 
+        /// </summary>
+        public static WhereEvaluator Instance { get; } = new WhereEvaluator();
```
IEvangelist, line 20:
```suggestion
        public static WhereEvaluator Instance { get; } = new();
```

src/Microsoft.Azure.CosmosRepository/Specification/ISpecification.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Paginate results, selects how many results should be returned
+        /// </summary>
+        int PageSize { get; }
+
+        /// <summary>
+        /// Use contiutation token instead of pagenumber
+        /// </summary>
+        bool UseContinutationToken { get; }
+
+
```
IEvangelist, line 50:
```suggestion
```

src/Microsoft.Azure.CosmosRepository/Specification/OrderByExpressionInfo.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.Azure.CosmosRepository.Specification
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class OrderExpressionInfo<T>
+    {
+        private readonly Lazy<Func<T, object>> keySelectorFunc;
```
IEvangelist, line 17:
```suggestion
        private readonly Lazy<Func<T, object>> _keySelectorFunc;
```

Naming violdation

src/Microsoft.Azure.CosmosRepository/Specification/OrderByExpressionInfo.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="orderType">Whether to (subsequently) sort ascending or descending.</param>
+        /// <exception cref="ArgumentNullException">If <paramref name="keySelector"/> is null.</exception>
+        public OrderExpressionInfo(Expression<Func<T, object>> keySelector, OrderTypeEnum orderType)
+        {
+            _ = keySelector ?? throw new ArgumentNullException(nameof(keySelector));
+
+            KeySelector = keySelector;
+            OrderType = orderType;
+
+            keySelectorFunc = new Lazy<Func<T, object>>(KeySelector.Compile);
```
IEvangelist, line 32:
```suggestion
            _keySelectorFunc = new Lazy<Func<T, object>>(KeySelector.Compile);
```

src/Microsoft.Azure.CosmosRepository/Specification/OrderByExpressionInfo.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Whether to (subsequently) sort ascending or descending.
+        /// </summary>
+        public OrderTypeEnum OrderType { get; }
+
+        /// <summary>
+        /// Compiled <see cref="KeySelector" />.
+        /// </summary>
+        public Func<T, object> KeySelectorFunc => keySelectorFunc.Value;
+
```
IEvangelist, line 49:
```suggestion
        public Func<T, object> KeySelectorFunc => _keySelectorFunc.Value;
```

src/Microsoft.Azure.CosmosRepository/Specification/Specificiations/BaseSpecification.cs
```diff
@@ -0,0 +1,56 @@
...
+            foreach(Expression<Func<T, bool>> filter in filters)
+            {
+                Query.Where(filter);
+            }
+        }
+        /// <inheritdoc/>
+        public IReadOnlyList<WhereExpressionInfo<T>> WhereExpressions { get; } = new List<WhereExpressionInfo<T>>();
+        /// <inheritdoc/>
+        public IReadOnlyList<OrderExpressionInfo<T>> OrderExpressions { get; } = new List<OrderExpressionInfo<T>>();
```
IEvangelist, line 42:
```suggestion
        public IReadOnlyList<WhereExpressionInfo<T>> WhereExpressions { get; } = new();
        /// <inheritdoc/>
        public IReadOnlyList<OrderExpressionInfo<T>> OrderExpressions { get; } = new();
```

src/Microsoft.Azure.CosmosRepository/Specification/Specificiations/ListSpecification.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Microsoft.Azure.CosmosRepository.Specification
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class ListSpecification<T> : BaseSpecification<T, IQueryResult<T>>
+        where T : IItem
+
```
IEvangelist, line 16:
```suggestion
```

IEvangelist (changes requested):
Lots of minor style nits, suggestions, and whatnot. I really like the look of this.

filipmhpersson:
> @filipmhpersson great work! Love the samples they look really cool, I think they're is just some XML comments to update and maybe some tests to add, and maybe a bit of formatting here and there but on the whole looking great.

Great I'm in the process I'm cleaning up and adding tests when there is time over. 

One thing I noticed that ordering with ThenBy/ThenByDescending requires a composite index set in the cosmos db. Do you think we should still make it possible for users to use ThenBy but try to make it explicit in comments and documentation that a composite index on the selected fields is necessary? Or is it better to remove the ThenBy functions and add it later if it's needed?

mumby0168:
> > @filipmhpersson great work! Love the samples they look really cool, I think they're is just some XML comments to update and maybe some tests to add, and maybe a bit of formatting here and there but on the whole looking great.
> 
> Great I'm in the process I'm cleaning up and adding tests when there is time over.
> 
> One thing I noticed that ordering with ThenBy/ThenByDescending requires a composite index set in the cosmos db. Do you think we should still make it possible for users to use ThenBy but try to make it explicit in comments and documentation that a composite index on the selected fields is necessary? Or is it better to remove the ThenBy functions and add it later if it's needed?

Yes I'd agree, I was curious whether the composite index would cause that isssu. I do believe it is documented by the Cosmos DB team that to use the ThenBy methods you do need a composite index.

filipmhpersson:
I have an issue I'm not really sure how to solve. Maybe some of you have more cosmos knowledge here. 

When doing a more complex query with continuation token I'm getting a result for a composite continuation token and not a "normal" token. It looks like this:

[{"compositeToken":{"token":"+RID:~uvMJAOhjU+aJAAAAAAAAAA==#RT:1#TRC:10#RTD:xr+kQyQkLTy4IjFZtUNUBMBFwA==#ISV:2#IEO:65567#QCF:4#FPC:AWUAAAAAAAAAyAAAAAAAAAA=","range":{"min":"","max":"FF"}},"orderByItems":[{"item":43}],"rid":"uvMJAOhjU+aJAAAAAAAAAA==","skipCount":0,"filter":"true"}]

It works perfectly fine for fetching items but it seems like this token cannot be used with a CountAsync. Then I get an exception that says :
MalformedContinuationTokenException: ParallelContinuationToken is missing field: 'token.

I guess this has something to do with the Count method in the cosmos SDK not expecting a composite token.

I'm not sure if the count even should include the continuation token but in the already existing PageAsync method with continuation token the count method uses the continutation token. So that's why I'm trying to make it work here. Another solution would be to just count without continuation token always. But I don't like that this would function different compared to the PageAsync method.

Any clues on how this could be solved? @mumby0168  @IEvangelist.

src/Microsoft.Azure.CosmosRepository/Specification/Builder/SpecificationBuilder.cs
filipmhpersson: Yes, these should be able to come in any order right? In Linq I could specify the order before there where clause

mumby0168:
> I have an issue I'm not really sure how to solve. Maybe some of you have more cosmos knowledge here. 
> 
> 
> 
> When doing a more complex query with continuation token I'm getting a result for a composite continuation token and not a "normal" token. It looks like this:
> 
> 
> 
> [{"compositeToken":{"token":"+RID:~uvMJAOhjU+aJAAAAAAAAAA==#RT:1#TRC:10#RTD:xr+kQyQkLTy4IjFZtUNUBMBFwA==#ISV:2#IEO:65567#QCF:4#FPC:AWUAAAAAAAAAyAAAAAAAAAA=","range":{"min":"","max":"FF"}},"orderByItems":[{"item":43}],"rid":"uvMJAOhjU+aJAAAAAAAAAA==","skipCount":0,"filter":"true"}]
> 
> 
> 
> It works perfectly fine for fetching items but it seems like this token cannot be used with a CountAsync. Then I get an exception that says :
> 
> MalformedContinuationTokenException: ParallelContinuationToken is missing field: 'token.
> 
> 
> 
> I guess this has something to do with the Count method in the cosmos SDK not expecting a composite token.
> 
> 
> 
> I'm not sure if the count even should include the continuation token but in the already existing PageAsync method with continuation token the count method uses the continutation token. So that's why I'm trying to make it work here. Another solution would be to just count without continuation token always. But I don't like that this would function different compared to the PageAsync method.
> 
> 
> 
> Any clues on how this could be solved? @mumby0168  @IEvangelist.

The continuation token won't work with the .CountAsync method. It only works if you perform the same query as you did when the token was generated. For example you can't have a query that orders by name and then by age with a continuation token. I wrote about this here if this helps explain 😃 

https://billy-mumby-dev.com/paging-in-azure-cosmos-db

I found this hard to begin with as well.

filipmhpersson:
> > I have an issue I'm not really sure how to solve. Maybe some of you have more cosmos knowledge here.
> > When doing a more complex query with continuation token I'm getting a result for a composite continuation token and not a "normal" token. It looks like this:
> > [{"compositeToken":{"token":"+RID:~uvMJAOhjU+aJAAAAAAAAAA==#RT:1#TRC:10#RTD:xr+kQyQkLTy4IjFZtUNUBMBFwA==#ISV:2#IEO:65567#QCF:4#FPC:AWUAAAAAAAAAyAAAAAAAAAA=","range":{"min":"","max":"FF"}},"orderByItems":[{"item":43}],"rid":"uvMJAOhjU+aJAAAAAAAAAA==","skipCount":0,"filter":"true"}]
> > It works perfectly fine for fetching items but it seems like this token cannot be used with a CountAsync. Then I get an exception that says :
> > MalformedContinuationTokenException: ParallelContinuationToken is missing field: 'token.
> > I guess this has something to do with the Count method in the cosmos SDK not expecting a composite token.
> > I'm not sure if the count even should include the continuation token but in the already existing PageAsync method with continuation token the count method uses the continutation token. So that's why I'm trying to make it work here. Another solution would be to just count without continuation token always. But I don't like that this would function different compared to the PageAsync method.
> > Any clues on how this could be solved? @mumby0168  @IEvangelist.
> 
> The continuation token won't work with the .CountAsync method. It only works if you perform the same query as you did when the token was generated. For example you can't have a query that orders by name and then by age with a continuation token. I wrote about this here if this helps explain 😃
> 
> https://billy-mumby-dev.com/paging-in-azure-cosmos-db
> 
> I found this hard to begin with as well.

So what would be the solution here? My currently pushed version works. But I do the Count with the query and without a continuation token. To me this makes sense, it's the total amount of items in the database given the current filter. But this will mean it's functions differently compare to the count of the PageAsync with continuation token method.

mumby0168:
I'll have a proper look 2 night @filipmhpersson away from my desk today.

mumby0168:
Hi, @filipmhpersson I have spoken with @IEvangelist and we have both agreed there is a great PR and we are happy for it to be merged.

We have some testing we'd like to do and some styling nits we will amend but other than we'd like to thank you for taking the time to raise this PR. If you are happy we can merge this and then get it released in the next few days :)

There are just some merge conflicts that need resolving first.

src/Microsoft.Azure.CosmosRepository/IQueryResult.cs
filipmhpersson: Ok will rename everywhere. Hopefully I'll get all place

> Hi, @filipmhpersson I have spoken with @IEvangelist and we have both agreed there is a great PR and we are happy for it to be merged.
> 
> We have some testing we'd like to do and some styling nits we will amend but other than we'd like to thank you for taking the time to raise this PR. If you are happy we can merge this and then get it released in the next few days :)
> 
> There are just some merge conflicts that need resolving first.

Awesome, I started yesterday with some styling (reming T to TItem where relevant). What else do you need from me? Should I resolve the merge conflicts and commit the styling I have done so far and you'll do what you need after that?

I have ideas on what could be done better but I think think this is a good enough version to get out there and see how it will work for developers. So it's probably better to just get this out and do improvements depending on feedback.

IEvangelist (approved):

mumby0168:
Are you happy for this to be merged @filipmhpersson ?

filipmhpersson:
> Are you happy for this to be merged @filipmhpersson ?

From my part I'm happy, but maybe some acceptance test and so is needed. Or is that handled later?

PR #172, opened 2022-01-13
Docs Attempt 2

mumby0168:
Trying to just see the `index.md` file displayed on the github pages site.


docs/_config.yml
```diff
@@ -1 +0,0 @@
-theme: jekyll-theme-hacker
```
IEvangelist, line 1: This one cannot be deleted as it is the theme I selected this AM

mumby0168: Added it back

IEvangelist (approved):

Issue #175, opened 2022-01-16
Specify some types in the same container but not others, is this possible?


WestDiscGolf:
**Describe the bug**
I've recently finished running through the [Pluralsight course on data modelling in cosmos db](https://app.pluralsight.com/library/courses/data-modeling-partitioning-patterns-azure-cosmos-db) and I thought it might be a fun exercise to see if I could implement the final container/type construction using the repository tooling. But I am unsure if either I'm not doing it right or it's not possible.

**To Reproduce**
I started off simple and worked through the meta data into the same container. For those which don't have access to the course it's essentially

Tag:
```
{
 "id" :"",
 "name" : "",
 "type": "tag"
}
```
Category:
```
{
 "id" :"",
 "name" : "",
 "type": "category"
}
```
Container: "meta"
Partition Key: "/type"

(not pasted an image from the course slide as unsure about copyright etc).

Reproduction code:

```C#
using Microsoft.Azure.CosmosRepository;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

const string databaseName = "datamodelling";

var host = Host.CreateDefaultBuilder(args)
    .ConfigureServices(services =>
    {
        services.AddCosmosRepository(options =>
        {
            // do not do this; demo only to make obvious
            options.CosmosConnectionString = "AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";

            options.DatabaseId = databaseName;
            options.ContainerPerItemType = false;

            options.ContainerBuilder.Configure<Category>(builder =>
            {
                builder.WithContainer("meta")
                    .WithPartitionKey("/type");
            });

            options.ContainerBuilder.Configure<Tag>(builder =>
            {
                builder.WithContainer("meta")
                    .WithPartitionKey("/type");
            });

        });
    })
    .Build();

var tagsRepository = host.Services.GetService<IRepository<Tag>>();
await tagsRepository.CreateAsync(new Tag() { Name = "production" });
await tagsRepository.CreateAsync(new Tag() { Name = "staging" });
await tagsRepository.CreateAsync(new Tag() { Name = "development" });

var categoryRepository = host.Services.GetService<IRepository<Category>>();
[...]

PR #179, opened 2022-01-17
Start on the partitioning strategy docs

mumby0168:
Basic explanation of how you can configure a partitioning strategy using the library. This was driven from the issue raised here #175.

There are still a few things I would like to add:

* Advice section, we set all partition keys on _all_ containers to `/partitionKey` this gives flexibility to change the PK if we decide it wasn't the best. It also allows _all_ items to be shared in a single container.
* Some some sort of FAQs or common errors and how to fix them, just some bits from personal experience.

### DIFF
--- a/docs/partitioning-setup.md
+++ b/docs/partitioning-setup.md
@@ -21,7 +21,7 @@ app.MapGet("/", () =>
 app.Run();
 ```
 
-The default partitioning strategy is simple all `IItem`'s will be in a single logical partition split by there unique IDs, which default to a GUID. This in small applications is often ample. A simple rule of thumb for this would be do you see yourself spanning over a physical partition? If no then this option _could_ be the way to go. However, in any scenarios where you are making an application that may grow in size, complexity or scale. It may be best to look at taking a bit more control over your partitioning strategy.
+The default partitioning strategy is simple, all `IItem`'s will be in a single logical partition split by there unique IDs. Which default to a GUID. This in small applications is often ample. A simple rule of thumb for whether or not this would be okay for you would be, do you see yourself spanning over a physical partition? If no then this option _could_ be the way to go. However, in any scenarios where you are making an application that may grow in size, complexity or scale. It may be best to look at taking a bit more control over your partitioning strategy.
 
 > This strategy is also demonstrated in the [`WebTier`](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/tree/main/samples/WebTier) sample application.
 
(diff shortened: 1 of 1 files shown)


docs/partitioning-setup.md
```diff
@@ -0,0 +1,146 @@
...
+
+app.MapGet("/", () => 
+    "Sample e-commerce Partitioning Strategy");
+
+app.Run();
+```
+
+#### Inventory Example
+
+The below example shows how two or more `IItem` types can share a container and a partitioning strategy. This define a `Stock` item which might hold the name, price, size and maybe weight of a piece of `Stock`. This is partitioned by the `/stockReferenceNumber`. The `StockRecord` item is a collection of `StockRecord`'s, this might be when a piece of stock was added or removed from the inventory, when this occurred and maybe a reason why. This is stored in the _same_ container as the `Stock`'s history that is recording. It also shares the same partitioning strategy. In the example below it means that all `Stock` and it's `StockRecord`'s are stored in the same logical partition in Cosmos DB.
```
mumby0168, line 90:
```suggestion
The below example shows how two or more `IItem` types can share a container and a partitioning strategy. This defines a `Stock` item that might hold the name, price, size, and maybe weight. This is partitioned by the `/stockReferenceNumber`. The `StockRecord` item is a collection of `StockRecord`'s. This record represents when a piece of stock was added or removed from the inventory, when this occurred, and maybe a reason why. This is stored in the _same_ container as the `Stock` item whose history it is recording. It also shares the same partitioning strategy. In the example below it means that all `Stock` and its `StockRecord`'s are stored in the same logical partition in Cosmos DB.
```

docs/partitioning-setup.md
```diff
@@ -0,0 +1,146 @@
+# Partitioning Setup
+
+Partitioning is a huge part of Cosmos DB. There are many different strategies you may want to take. Such as, splitting different types of item's across containers, sharing some `IItem`'s in one container and not others and finally, in some cases, even putting all `IItem` types in a single container.
+
+## Default Strategy
+
+Adding the Azure Cosmos Repository with no real configuration, like below. Will result in a simple strategy being assumed on your behalf. This will leave the option's configured in there default state. This means a property called `ContainerPerItemType` is set to false. This means that the Azure Cosmos Repository will store _all_ `IItem`'s in the same container, setting all the `IItem`'s within to use the partition key `/id`.
+
+```c#
```
IEvangelist, line 9:
```suggestion
```csharp
```

IEvangelist (approved):

IEvangelist (approved):

PR #190, opened 2022-01-18
VS'22 and added support for serverless resources, null throughput

IEvangelist:
Hey @mumby0168 - this is what we were talking about. I think it's what we'd need, have a look.

### DIFF
--- a/tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
+++ b/tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
@@ -73,5 +73,16 @@ namespace Microsoft.Azure.CosmosRepositoryTests.Providers
             Assert.Contains("500", exception.Message);
             Assert.Contains("700", exception.Message);
         }
+
+        [Fact]
+        public void GetThroughputPropertiesContainerWithServerlessConfiguration()
+        {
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
```diff
@@ -73,5 +73,16 @@ public void GetThroughputPropertiesItemsWithConflictingManualValues()
...
             Assert.Contains("700", exception.Message);
         }
+
+        [Fact]
+        public void GetThroughputPropertiesContainerWithServerlessConfiguration()
+        {
+            _repositoryOptions.ContainerBuilder.Configure<AnotherTestItem>(
+                builder => builder.WithServerlessThroughput());
+
+            ThroughputProperties throughputProperties = _provider.GetThroughputProperties<AnotherTestItem>();
```
mumby0168, line 83: Do you know if we can enable nullable reference types even tho we are on netstandard 2.0? Not one for this PR but was just curious?

mumby0168 (approved):

tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
IEvangelist: Yes, NRT is fully supported in .NET Standard 2.0

PR #192, opened 2022-01-20
GetAsync<T> Optional Type Checking

### LINKED ISSUE #152: Cannot use GetAsync<T>() to retrieve subclasses of T
**Describe the bug**
Cannot retrieve subclass types.

**To Reproduce**
* My repository is a `DefaultRepository<Offering>`.
* My project has two subclasses: `Document : Offering` `Module : Offering`.
* When storing them I can see that the `type` property in Cosmos DB is set to `Document`, respectively `Module`.
* I'm using `GetAsync()` to retrieve an `Offering`, however I'm always getting NULL back because `GetAsync()` refuses to return an instance if the `type` property's value does not match (https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/Microsoft.Azure.CosmosRepository/src/DefaultRepository.cs#L79).

**Expected behavior**
Return an instance of the subtype.

**Actual behavior**
Returns NULL.

**Environment summary**
SDK Version: 2.8.0
OS Version: Windows 11

**Additional context**
I get the idea of a repository being meant for one specific type. However, there's something odd about the logic, I think. If I can **save** any type to my `IRepository<T>`, I should also be able to retrieve it again. A workaround for me would be to have two repositories, one for each specific sub type. However, that would break my logic to (for example) get all offerings of a specific date, which would be a mixed list of documents and modules.

To overcome the limitation, I also tried to implement my own `IRepository<T>` instead of using `DefaultRepository<T>`, so I could add sub type support. However, as most of the classes are internal, it's not possible to implement your own repo, unless I'm missing something?

mumby0168:
Closes #152


tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/TypeCheckingTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        await buyOneGetOneFreeOffersRepository.CreateAsync(buyOneGetOneFreeOffer);
+        await buyOneGetOneHalfPriceRepository.CreateAsync(buyOneGetOneHalfPriceOffer);
+
+        //Act
+        IEnumerable<Offer> results = await offersRepository.GetAsync(x =>
+            x.OfferType == nameof(BuyOneGetOneFreeOffer) &&
+            x.OfferType == nameof(BuyOneGetOneHalfPriceOffer));
+
+        //Assert
+        results.Count().Should().Be(2);
```
Krumelur, line 70: @mumby0168 Not only should this return two items but `results[0].GetType() == typeof(BuyOneGetOneFreeOffer)` and `results[1].GetType() == typeof(BuyOneGetOneHalfPriceOffer)`.

mumby0168: Not sure I am following that bit? `GetType()` is not support by Cosmos DB when the query is translated to SQL?

Krumelur: This was meant for the test. What I'm saying is: if you store a `BuyOneGetOneHalfPriceOffer` and you read an `Offer` (`Offer offer = Get("id_of_subtype_of_offer")`), `offer` must be a `BuyOneGetOneHalfPriceOffer` and not `Offer`.

IEvangelist (dismissed):
This looks good to me, thanks @mumby0168

tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/TypeCheckingTests.cs
mumby0168: So you want to try and read a BuyOneGetOneHalfPriceOffer using an `IRepository<Offer>` something like
`repository.GetAsync(buyOneGetOneHalfPriceOfferId);` ?

Krumelur:
Exactly. If you store different subtypes in the same container, you will want to get all matching a specific query, irrespective of what type they are. 

Scroll back up to see how i solved this using `JsonSubTypes`. However, this only works as long as all types are returned. 

René

mumby0168:
@Krumelur we could look at setting the Type = "" or disable the type checking optionally via configuration at startup time.

tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/DisabledStrictTypeCheckingTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        options.DatabaseId = OffersDatabaseName;
+        options.CosmosConnectionString = GetCosmosConnectionString();
+        options.ContainerPerItemType = true;
+
+        options.ContainerBuilder.Configure<DisabledOffer>(builder =>
+        {
+            builder.WithContainer(OffersContainerName);
+            builder.WithPartitionKey(OffersPartitionKeyPath);
+            builder.WithContainerDefaultTimeToLive(TimeSpan.FromMinutes(10));
+            builder.WithoutStrictTypeChecking();
```
mumby0168, line 47: @Krumelur this is the proposal, you can turn this off to not force the type to be checked. However, this is at your own risk bearing in mind that the sub type MUST deserialize into the base type.

@IEvangelist what do you think? I have had this use case as well in the past and think for base types such as this. The ability to disable this feature albeit at your own risk could be very useful.

IEvangelist (approved):

tests/Microsoft.Azure.CosmosRepositoryAcceptanceTests/DisabledStrictTypeCheckingTests.cs
IEvangelist: I'm good with that, thanks for handling this.

IEvangelist (approved):

Krumelur:
> we could look at setting the Type = "" or disable the type checking optionally via configuration at startup time.

I don't follow. I thought this issue and the PR was about disabling the Type checking because it is what causes the problem? I'm puzzled.

mumby0168:
> > we could look at setting the Type = "" or disable the type checking optionally via configuration at startup time.
> 
> I don't follow. I thought this issue and the PR was about disabling the Type checking because it is what causes the problem? I'm puzzled.

Yes that is exactly what we have done. Given your scenario, with the base `Offering` class you can tell the repository to NOT perform strict type checking. Something like this.

```csharp
options.ContainerBuilder.Configure<Offering>(builder =>
{
    builder.WithContainer(OffersContainerName);
    builder.WithPartitionKey(OffersPartitionKeyPath);
    builder.WithContainerDefaultTimeToLive(TimeSpan.FromMinutes(10));
    builder.WithoutStrictTypeChecking();
});
```

> Note the `builder.WithoutStrictTypeChecking();`

Does this make sense? The strict type checking is on for every item as a default as this is how it has always worked. But you have the option to turn it off.

I am going to release a pre-release package, would you like to try it out?

sven5:
@mumby0168 
There has been silence for a while now. 😄 
I was able to confirm that it works. Thanks.

PR #196, opened 2022-01-21
Update tests to run UT as part of a release

mumby0168:


mumby0168:
Thoughts on running acceptance tests prior to a release @IEvangelist?

I am not sure if we need to see them stabilize a little first?

IEvangelist:
> Thoughts on running acceptance tests prior to a release @IEvangelist?
> 
> I am not sure if we need to see them stabilize a little first?

Yes, I think that is a good idea. If we have the resources to do so, let's do it.

IEvangelist (dismissed):

IEvangelist (approved):

PR #200, opened 2022-01-30
Specification tweaks

mumby0168:


.github/workflows/nuget-publish.yml
```diff
@@ -23,7 +23,24 @@ jobs:
         
         - name: Unit Tests
           run: |
-            dotnet test --no-restore --verbosity normal --filter "Category!=Acceptance"
+            dotnet test --no-restore --verbosity normal --filter "Category!=Acceptance"        
+            
+    ac_tests:
```
IEvangelist, line 28:
```suggestion
    acceptance_tests:
```

.github/workflows/nuget-publish.yml
```diff
@@ -50,7 +67,9 @@ jobs:
             dotnet test --no-restore --verbosity normal --filter "Category=Acceptance&Type=Container"
 
     deploy:
-      needs: test
+      needs:
+        - ac_tests
```
IEvangelist, line 71:
```suggestion
        - acceptance_tests
```

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
```diff
@@ -29,20 +28,17 @@ internal class InMemoryRepository<TItem> : IRepository<TItem> where TItem : clas
     {
         internal long CurrentTs => DateTimeOffset.UtcNow.ToUnixTimeSeconds();
         internal ConcurrentDictionary<string, string> Items { get; } = new();
-        internal ISpecificationEvaluator _specificationEvaluator { get; set; }
+
+        private readonly ISpecificationEvaluator _specificationEvaluator;
```
IEvangelist, line 32: For private fields, let's try to keep those at the top of the file, then internals.

src/Microsoft.Azure.CosmosRepository/Specification/Builder/SpecificationBuilderExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <typeparam name="TItem"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <param name="specificationBuilder"></param>
+        /// <param name="criteria"></param>
+        public static ISpecificationBuilder<TItem, TResult> Where<TItem, TResult>(
+            this ISpecificationBuilder<TItem, TResult> specificationBuilder,
+            Expression<Func<TItem, bool>> criteria)
+            where TItem : IItem
+            where TResult : IQueryResult<TItem>
+
```
IEvangelist, line 27:
```suggestion
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/IEvaluator.cs
```diff
@@ -29,6 +26,5 @@ public interface IEvaluator
         IQueryable<TItem> GetQuery<TItem, TResult>(IQueryable<TItem> query, ISpecification<TItem, TResult> specification)
```
IEvangelist, line 26:
```suggestion
        IQueryable<TItem> GetQuery<TItem, TResult>(
            IQueryable<TItem> query,
            ISpecification<TItem, TResult> specification)
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/OrderEvaluator.cs
```diff
@@ -8,46 +8,55 @@
...
+    internal class OrderEvaluator : IEvaluator
     {
-        private OrderEvaluator() { }
-        public static OrderEvaluator Instance { get; } = new OrderEvaluator();
+        public bool IsFilterEvaluator => false;
 
-        public bool IsFilterEvaluator { get; } = false;
-
-        public IQueryable<TItem> GetQuery<TItem, TResult>(IQueryable<TItem> query, ISpecification<TItem, TResult> specification)
+        public IQueryable<TItem> GetQuery<TItem, TResult>(IQueryable<TItem> query,
```
IEvangelist, line 15:
```suggestion
        public IQueryable<TItem> GetQuery<TItem, TResult>(
            IQueryable<TItem> query,
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/PagingEvaluator.cs
```diff
@@ -1,52 +1,32 @@
...
-        /// </summary>
-        /// <typeparam name="TItem"></typeparam>
-        /// <typeparam name="TResult"></typeparam>
-        /// <param name="query"></param>
-        /// <param name="specification"></param>
-        /// <returns></returns>
-        public IQueryable<TItem> GetQuery<TItem, TResult>(IQueryable<TItem> query, ISpecification<TItem,TResult> specification)
+        public bool IsFilterEvaluator => false;
+
+        public IQueryable<TItem> GetQuery<TItem, TResult>(IQueryable<TItem> query,
```
IEvangelist, line 12:
```suggestion
        public IQueryable<TItem> GetQuery<TItem, TResult>(
            IQueryable<TItem> query,
```

src/Microsoft.Azure.CosmosRepository/Specification/Evaluator/WhereEvaluator.cs
```diff
@@ -8,38 +8,13 @@
...
-        {
-            foreach (WhereExpressionInfo<T> info in specification.WhereExpressions)
-            {
-                query = query.Where(info.Filter);
-            }
+        public bool IsFilterEvaluator => true;
 
-            return query;
-        }
+        public IQueryable<T> GetQuery<T, TResult>(IQueryable<T> query, ISpecification<T, TResult> specification)
```
IEvangelist, line 15:
```suggestion
        public IQueryable<T> GetQuery<T, TResult>(
            IQueryable<T> query,
            ISpecification<T, TResult> specification)
```

src/Microsoft.Azure.CosmosRepository/Specification/ISpecification.cs
```diff
@@ -19,20 +15,24 @@ public interface ISpecification<TItem, out TResult>
         /// <summary>
         /// The collection of filter expressions used for filtering queries.
         /// </summary>
-        IReadOnlyList<WhereExpressionInfo<TItem>> WhereExpressions { get; }
+        List<WhereExpressionInfo<TItem>> WhereExpressions { get; }
```
IEvangelist, line 18: I'm curious why the change from an `IReadOnlyList` to a `List`? This exposes the ability for consumers to call `.Add`, is that in fact desired?

IEvangelist (approved):
This is looking really sharp! Love it. I have some nits, but really like this. I also have a few questions.

filipmhpersson:
> I need to do some more reading of this code to wrap my head around it properly.
> 
> I have been looking at this https://ardalis.github.io/Specification/usage/use-specification-repository-pattern.html as this is what the implementation was based on it seems.
> 
> I have a few points that I was wondering your thoughts on @IEvangelist @filipmhpersson
> 
> 1. What do we _need_ to expose in terms of the builders and the direct access to the `Specification` object. I was thinking of encapsulating most of this. But does this limit extensiblity?
> 2. What is the benefit of having the extension methods for each builder rather than just having those defined on the builder.
> 3. Do we look at limiting the calls a user can make. For example, you cannot do `Query.OrderBy()` before calling `.Where` I feel we should enforce this by the path we take with the fluent API.
> 
> Todo:
> 
> * [ ]  Unit test the specifications & their builders
> * [x]  Acceptance tests on filtering
> * [x]  Acceptance tests on CT paging with filters
> * [ ]  Acceptance tests for skip, take paging approach
> * [ ]  Ensure logging is correct
> * [ ]  Basic docs on this new feature

1: In general many things can be encapsulated here. I wasn't sure how you wanted to go and therefor didn't do it in my PR. The intended use for developers is to implement new specifications and use the builder to build those specifications, I see no need to give direct access to the specification object. They shouldn't interact with other parts of the specification directly. The builder is how we limit what's possible to do with the specification. 

2: Don't see a reason why it can't be done both ways.
3: If you feel that this is a better experience for the developer go for it. One thing that also would be really nice to limit is to hide the ContinuationToken and PageNumber methods completely until you base the implementations on those specific specifications.

mumby0168:
@filipmhpersson Hi, I also wanted you to cast your eyes over the acceptance tests do you think they are suitable?

Okay, I think we can tie this implementation down a lot given what you have said. I think we can start by tieing as much down as possible. Then if we find it hampers extensibility. We can simply expose it later.

As for the extensions methods. I think these may better sit on the builders themselves. That is just a personal preference. Not sure if you ever would want to mock them. But if you did the extension methods would not play nice.

As for the limiting of the `.OrderBy` the tests show this actually works without a `.Where` clause. This is because we add this into the library ourselves for the `Type` field. I think we can do some tricks with interfaces to only allow the `.ThenBy` if an `.OrderBy` has been called.

Really great contribution, thanks again.

src/Microsoft.Azure.CosmosRepository/InMemoryRepository.cs
mumby0168: Will ammend.

src/Microsoft.Azure.CosmosRepository/Specification/ISpecification.cs
mumby0168: I did this for now until I understood whether or not this interface should even be `public` but I am coming to think it should not be.

mumby0168:
Going to try and get some docs written.

src/Microsoft.Azure.CosmosRepository/Specification/ISpecification.cs
mumby0168: Updated this. Back to `IReadOnlyList<T>`/

IEvangelist (approved):

PR #202, opened 2022-01-30
Microsoft.Azure.CosmosEventSourcing

mumby0168:


src/Microsoft.Azure.CosmosEventSourcing/ChangeFeed/EventSourcingProcessor.cs
```diff
@@ -0,0 +1,101 @@
...
+            catch (Exception e)
+            {
+                _logger.LogError(e,
+                    "Failed handling projection for container {ContainerName} source event ID {SourcedEventId}",
+                    containerName, change.Id);
+            }
+        }
+    }
+
+    private Task OnErrorAsync(Exception exception, string containerName)
```
IEvangelist, line 90: Should we also broadcast errors for consumers, who might want (or need) to take a different action if something goes wrong?

src/Microsoft.Azure.CosmosEventSourcing/EventSourcingRepository.cs
```diff
@@ -0,0 +1,28 @@
...
+
+internal class EventSourcingRepository<TEventStream> : IEventSourcingRepository<TEventStream> where TEventStream : SourcedEvent
+{
+    private readonly IRepository<TEventStream> _repository;
+
+    public EventSourcingRepository(IRepository<TEventStream> repository) =>
+        _repository = repository;
+
+    public async ValueTask PersistAsync(TEventStream record) =>
+        await _repository.CreateAsync(record);
```
IEvangelist, line 16:
```suggestion
    public ValueTask PersistAsync(TEventStream record) =>
        _repository.CreateAsync(record);
```

src/Microsoft.Azure.CosmosEventSourcing/EventSourcingRepository.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public async ValueTask PersistAsync(IEnumerable<TEventStream> records) =>
+        await _repository.CreateAsync(records);
```
IEvangelist, line 19:
```suggestion
    public ValueTask PersistAsync(IEnumerable<TEventStream> records) =>
        _repository.CreateAsync(records);
```

IEvangelist (commented):
I'll have to read up on this a bit more. It reads really well, no major concerns thus far. I'll look a bit deeper this week.

src/Microsoft.Azure.CosmosEventSourcing/ChangeFeed/EventSourcingProcessor.cs
mumby0168: Yes, I like that idea. I am not sure what triggers this as there is no reference to a specific change. I will need to read the docs.

mumby0168:
> I'll have to read up on this a bit more. It reads really well, no major concerns thus far. I'll look a bit deeper this week.

Okay, that is great! I need to work on actually adding some meaningful projections for the event data into the sample. There is also a concept of snapshots that I have not got my head around yet. This is for cases where there are large event sets.

It feels like a good start tho I think Cosmos DB is a great fit for this design pattern.

src/Microsoft.Azure.CosmosEventSourcing/Converters/PersistedEventConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) =>
+        serializer.Serialize(writer, value, value.GetType());
+
+    public override object ReadJson(
+        JsonReader reader,
+        Type objectType,
+        object existingValue,
+        JsonSerializer serializer)
+    {
+        JToken? j = JToken.ReadFrom(reader);
```
mumby0168, line 23: @IEvangelist do you know a better way to do this? 

I'd be happy to use System.Text if it makes it easier.

mumby0168: Okay so been trying a few things - this seems to be the most efficient

https://github.com/mumby0168/PolymorphicDotnetJson/blob/b98250ddede01fcf69cb257a64c6f6615a87986f/src/PolymorphicDotnetJson/SystemText/SystemTextEventConverter.cs#L6

That is baed on these benchmarks

The first uses what I have now with a `JObject` the other uses a `JsonDocument` and the final one uses a raw parse of the JSON which I thought would be better but appears not.

Screenshot 2022-02-05 at 17 22 44

dcuccia: This is exciting! I'm not well versed in event sourcing architectures, but just a high-level comment that if we're looking for inspiration from other impl's in document dbs, there's Marten by Jeremy Miller: https://github.com/JasperFx/marten

mumby0168: Ahhh yeah that's very cool! I'd heard of this before, but I haven't properly looked at it. Thanks @dcuccia

samples/BasicEventSourcingSample/Projections/ShipInformationProjections.cs
```diff
@@ -0,0 +1,80 @@
...
+                nameof(ShipInformation),
+                cancellationToken);
+
+            shipInfo.LatestPort = port;
+
+            await _repository.UpdateAsync(shipInfo, cancellationToken);
+        }
+    }
+
+    public class ShipLoadedHandler : IEventProjectionHandler<ShipEvents.Loaded>
```
IEvangelist, line 56: In looking at these handlers, this could become pretty repetitive.  And while it does offer the most flexibility, and I we should expose this to the consumers, it would be really nice to do some of that work for the consumer. For example, imagine being able to register just the handler action:

```csharp
// They would call MapEventProjectionHandler, given the pairing of 
// ShipInformation and the Loaded event.
// They would then provide the handler delegate, which is lines 66-78.
services.MapEventProjectionHandler<ShipInformation, ShipEvents.Loaded>(
    handler: Func<IRepository<ShipInformation> repo,
                  ShipEvents.Loaded loaded,
                  CancellationToken cancellationToken,
                  ValueTask>);
```

When they map a handler in this way, we'd have to detect that and do the work of pairing the types and doing the DI for them.

```csharp
class DefaultEventProjectionHandler<TItem, TEvent> : 
    IEventProjectionHandler<TEvent> where TEvent : IPersistedEvent
    where TItem : IItem
{
    readonly IRepository<TItem> _repositoryOfT;
    readonly IEventProjectionHandlerStore _handlerStore;

    public DefaultEventProjectionHandler(
        IRepository<TItem> repositoryOfT,
        IEventProjectionHandlerStore handlerStore) => 
        (_repositoryOfT, _handlerStore) = (repositoryOfT, handlerStore);

    ValueTask IEventProjectionHandler<TEvent>.HandleAsync(
        TEvent @event,
        CancellationToken cancellationToken = default) =>
        _handlerStore.MappedFrom<TItem, TEvent>(@event, cancellationToken);
}
```

Something similar to that, this is of course pulled from my imagination so the implementation details would likely vary. The `IEventProjectionHandlerStore` would have a thread-safe, generic-type collection lookup of mapped handlers. Does that make sense?

src/Microsoft.Azure.CosmosEventSourcing/ChangeFeed/EventSourcingProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        _logger.LogError(exception, "Failed handling when handling changes detected from container {ContainerName}",
+            containerName);
+        return Task.CompletedTask;
+    }
+
+    public Task StopAsync() =>
+        _processor?.StopAsync() ?? Task.CompletedTask;
+
+    public IReadOnlyList<Type> ItemTypes => new[] {typeof(TSourcedEvent)};
```
IEvangelist, line 96:
```suggestion
    public IReadOnlyList<Type> ItemTypes => new[] { typeof(TSourcedEvent) };
```

src/Microsoft.Azure.CosmosEventSourcing/ChangeFeed/EventSourcingProvider.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosRepository.ChangeFeed;
+using Microsoft.Azure.CosmosRepository.ChangeFeed.Providers;
+
+namespace Microsoft.Azure.CosmosEventSourcing.ChangeFeed;
+
+public class EventSourcingProvider : IChangeFeedContainerProcessorProvider
```
IEvangelist, line 9:
```suggestion
internal class EventSourcingProvider : IChangeFeedContainerProcessorProvider
```

Does the implementation need to be public?

src/Microsoft.Azure.CosmosEventSourcing/EventSourcingProcessorOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.CosmosEventSourcing;
+
+public class EventSourcingProcessorOptions<TSourceEvent> where TSourceEvent : EventSource
+{
+    /// <summary>
+    /// The instance name of the processor.
+    /// </summary>
+    public string InstanceName { get; set; } = "default";
```
IEvangelist, line 11:
```suggestion
    public string InstanceName { get; set; } = "default-processor-instance-name";
```

src/Microsoft.Azure.CosmosEventSourcing/Extensions/Extensions.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            assemblies = AppDomain.CurrentDomain.GetAssemblies();
+        }
+
+        List<Type> types = assemblies
+            .SelectMany(x => x.GetTypes()
+                .Where(type => type.IsAssignableTo(typeof(IPersistedEvent))))
+            .ToList();
+
+        types.ForEach(x => PersistedEventConverter.ConvertableTypes.Add(x));
```
IEvangelist, line 85:
```suggestion
        types.ForEach(PersistedEventConverter.ConvertableTypes.Add);
```

IEvangelist (commented):
A few more nits while having a look, and a few ideas.

samples/BasicEventSourcingSample/Core/Aggregate.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License.
+
+using System.Reflection.Metadata;
+using System.Security.Policy;
+using Microsoft.Azure.CosmosEventSourcing;
+
+namespace BasicEventSourcingSample.Core;
+public abstract class Aggregate
+{
+    private List<IPersistedEvent> _events = new();
```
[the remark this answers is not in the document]
mumby0168: I think you are right we can scrap this.

src/Microsoft.Azure.CosmosEventSourcing/Projections/EventBasedSourceProjectionBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+            _logger.LogDebug("No IEventProjectionHandler<{EventType}> found",
+                payloadTypeName);
+            return;
+        }
+
+        await Task.WhenAll(handlers.Select(handler =>
+        {
+            try
+            {
+                object? result = handlerType.GetMethod("HandleAsync")?
```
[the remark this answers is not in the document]
mumby0168: I tried this but unless we have a default `IPersistedEvent` we can't, I can add one however.

samples/BasicEventSourcingSample/Program.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    containerBuilder.Configure<ShipInformation>(shipContainerOptions =>
+    {
+        shipContainerOptions.WithContainer("ships-projections");
+        shipContainerOptions.WithPartitionKey("/partitionKey");
+    });
+
+    containerBuilder.ConfigureEventSourceStore<ShipEventSource>("ship-tracking-events");
+});
+
```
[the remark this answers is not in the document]
mumby0168: Thanks 😃

src/Microsoft.Azure.CosmosEventSourcing/Projections/EventBasedSourceProjectionBuilder.cs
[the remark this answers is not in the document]
IEvangelist: You should be able to use the generic-type parameter `TEventSource` which appears to be in scope. There shouldn't be a need to create a default one.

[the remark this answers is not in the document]
mumby0168: Ah I looked at this and the `TEventSource` isn't a `IPersistedEvent` in this case so the compiler doesn't like it.

samples/BasicEventSourcingSample/Core/ShipApplys.cs
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using CleanArchitecture.Exceptions;
+using Microsoft.Azure.CosmosEventSourcing;
+
+namespace BasicEventSourcingSample.Core;
+
+public partial class Ship
```
IEvangelist, line 9: I renamed this file to _Ship.Apply.cs_, singular and use the `.` to clearly separate the `Ship` and it's partial implementations specific intent.

samples/BasicEventSourcingSample/Infrastructure/ShipRepository.cs
```diff
@@ -0,0 +1,61 @@
...
+            .GetAsync(x => x.PartitionKey == nameof(ShipInformation));
+
+        return all.Select(x => x.Name);
+    }
+
+    public async ValueTask<ShipInformation?> GetInformationAsync(string name)
+    {
+        try
+        {
+            return await _shipInformationRepository.GetAsync(name, nameof(ShipInformation));
```
IEvangelist, line 54: This would be a nice use-case for the newly proposed `TryGetAsync` API we were talking about, right - or am I thinking of something else?

samples/BasicEventSourcingSample/Projections/ShipInformationProjections.cs
IEvangelist: Once this is merged, I'd like to explore this a bit more too. I think it could be pretty slick. Note to future self, investigate this.

src/Microsoft.Azure.CosmosEventSourcing/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <param name="services">The collection used to add the services for the cosmos event sourcing library.</param>
+    /// <param name="eventSourcingBuilder">A <see cref="ICosmosEventSourcingBuilder"/> used to configure the library.</param>
+    /// <returns>The instance of the <see cref="IServiceCollection"/></returns>
+    public static IServiceCollection AddCosmosEventSourcing(
+        this IServiceCollection services,
+        Action<ICosmosEventSourcingBuilder> eventSourcingBuilder)
+    {
+        CosmosEventSourcingBuilder builder = new(services);
+        eventSourcingBuilder.Invoke(builder);
+        services.AddSingleton(typeof(IEventSourceRepository<>), typeof(EventSourceRepository<>));
```
IEvangelist, line 28: Thinking for consistency we might want to continue to use the `Default` name as part of these newly added members. For example, `DefaultEventSourceRepository` and `DefaultEventSourcingProvider`.

IEvangelist (approved):
A few minor nits and thoughts, very well done! This is beautiful!!

src/Microsoft.Azure.CosmosEventSourcing/Extensions/ServiceCollectionExtensions.cs
mumby0168: I like this updated.

samples/BasicEventSourcingSample/Infrastructure/ShipRepository.cs
mumby0168: Yeah, that's the exact use case, it will be really nice.

samples/BasicEventSourcingSample/Core/ShipApplys.cs
mumby0168: Yeah, I like this. It makes sense.

IEvangelist (approved):

PR #213, opened 2022-03-02
Update paging methods to be opt in for returning count information

robertbennett1998:
Currently any paging requests (in this package) also make a count request, this can be very expensive to run and therefore shouldn't be ran be default but should be opt in.

https://stackoverflow.com/questions/27745346/get-record-count-in-azure-documentdb
> For example from the above thread: 
16,400 records in single partition consumes 535 RUs.  
[Youngjae](https://stackoverflow.com/users/361100/youngjae)
 [Jul 24, 2017 at 4:57](https://stackoverflow.com/questions/27745346/get-record-count-in-azure-documentdb#comment77508424_43560975)

# Breaking Changes
- Added default variable to opt out (by default) or returning the count information from paging queries.
    - This will require users who wish to get total count information to add an additional parameter to their calls.

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
+++ b/src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
@@ -370,6 +370,7 @@ namespace Microsoft.Azure.CosmosRepository
             Expression<Func<TItem, bool>>? predicate = null,
             int pageSize = 25,
             string? continuationToken = null,
+            bool returnTotal = false,
             CancellationToken cancellationToken = default)
         {
             Container container = await _containerProvider.GetContainerAsync().ConfigureAwait(false);
@@ -379,7 +380,6 @@ namespace Microsoft.Azure.CosmosRepository
                 MaxItemCount = pageSize
             };
 
-
             IQueryable<TItem> query = container
                 .GetItemLinqQueryable<TItem>(requestOptions: options, continuationToken: continuationToken)
                 .Where(_repositoryExpressionProvider.Build(predicate ??
@@ -387,7 +387,11 @@ namespace Microsoft.Azure.CosmosRepository
 
             _logger.LogQueryConstructed(query);
 
-            Response<int> countResponse = await query.CountAsync(cancellationToken);
+            Response<int>? countResponse = null;
+            if (returnTotal)
+            {
+                countResponse = await query.CountAsync(cancellationToken);
+            }
 
             (List<TItem> Items, double Charge, string? ContinuationToken) result =
                 await GetAllItemsAsync(query, pageSize, cancellationToken);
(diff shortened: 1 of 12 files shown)


src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -379,26 +380,29 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
                                                            _repositoryExpressionProvider.Default<TItem>()));
 
             _logger.LogQueryConstructed(query);
 
-            Response<int> countResponse = await query.CountAsync(cancellationToken);
+            Response<int>? countResponse = null;
+            if (returnTotal)
+            {
+                countResponse = await query.CountAsync(cancellationToken);
+            }
```
mumby0168, line 394:
```suggestion
            Response<int>? countResponse = null;
            
            if (returnTotal)
            {
                countResponse = await query.CountAsync(cancellationToken);
            }
```

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -379,26 +380,29 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
...
 
             _logger.LogQueryExecuted(query, result.Charge);
 
             return new Page<TItem>(
-                countResponse.Resource,
+                countResponse?.Resource ?? null,
                 pageSize,
                 result.Items.AsReadOnly(),
-                result.Charge + countResponse.RequestCharge,
+                result.Charge + countResponse?.RequestCharge ?? 0,
```
mumby0168, line 405: Does this still need to sum up the charge from the count and the query?

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -416,8 +421,11 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
                 .Where(_repositoryExpressionProvider
                     .Build(predicate ?? _repositoryExpressionProvider.Default<TItem>()));
 
-            Response<int> countResponse =
-                await query.CountAsync(cancellationToken).ConfigureAwait(false);
+            Response<int>? countResponse = null;
```
mumby0168, line 424:
```suggestion
            Response<int>? countResponse = null;
            
```

src/Microsoft.Azure.CosmosRepository/DefaultRepository.cs
```diff
@@ -430,11 +438,11 @@ public async ValueTask<int> CountAsync(CancellationToken cancellationToken = def
             _logger.LogQueryExecuted(query, result.Charge);
 
             return new PageQueryResult<TItem>(
-                countResponse.Resource,
+                countResponse?.Resource ?? null,
                 pageNumber,
                 pageSize,
                 result.Items.AsReadOnly(),
-                result.Charge + countResponse.RequestCharge);
+                result.Charge + countResponse?.RequestCharge ?? 0);
```
mumby0168, line 445: Add charge again I think.

mumby0168 (approved):
Looks good to me. I much prefer this to be an opt-in feature given the potential cost implications on large data sets. @IEvangelist are you cool with this?

IEvangelist (approved):

PR #214, opened 2022-03-02
Allow v3 releases

mumby0168:


.github/workflows/nuget-publish.yml
```diff
@@ -2,7 +2,7 @@ name: Release Package Version
 on:
   push:
     tags:
-     - 2.*     
+     - 3.*
```
IEvangelist, line 5:
```suggestion
     - *.*     
```

I wonder if we don't just make it valid for any tag?

IEvangelist (approved):

.github/workflows/nuget-publish.yml
mumby0168: Yeah, that should work.

IEvangelist (approved):

PR #219, opened 2022-03-05
Renaming, few extra methods, extra data available on projections.

mumby0168:
## Added EventSource to the projection builder

There may be common data stored on an `EventSource` that might not be on every `IPersistedEvent` which a user may want access to. This also allows access to the `Id` on the `EventSource` which could be used in an event inbox to stop duplicates being processed for example.

## Renamed `IEventSourceRepository<T>` -->  `IEventStore<T>`

I felt this better describes this as just a store of events and not an `IRepository<T>` as I feel you should always have a single container per set of events you are storing.

## Extra method to filter events

This could be useful pulling reports etc being able to find all of a given event or all of a set of events.

### DIFF
--- a/src/Microsoft.Azure.CosmosEventSourcing/Extensions/EventSourceExtensions.cs
+++ b/src/Microsoft.Azure.CosmosEventSourcing/Extensions/EventSourceExtensions.cs
@@ -26,8 +26,8 @@ public static class EventSourceExtensions
     /// <returns>The TEvent instance.</returns>
     /// <remarks>If the event payload cannot be converted to the TEvent type null is returned.</remarks>
     public static TEvent? TryGetEventPayload<TEvent>(this EventSource eventSource)
-        where TEvent : class, IPersistedEvent
-        => eventSource.EventPayload switch
+        where TEvent : class, IPersistedEvent =>
+        eventSource.EventPayload switch
         {
             TEvent eventPayload => eventPayload,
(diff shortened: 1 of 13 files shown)


src/Microsoft.Azure.CosmosEventSourcing/Extensions/EventSourceExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Cast's the payload of an <see cref="EventSource"/> to a <see cref="IPersistedEvent"/>
+    /// </summary>
+    /// <param name="eventSource">The <see cref="EventSource"/> to read the payload from.</param>
+    /// <typeparam name="TEvent">The event type the payload will be cast to.</typeparam>
+    /// <returns>The TEvent instance.</returns>
+    /// <remarks>If the event payload cannot be converted to the TEvent type null is returned.</remarks>
+    public static TEvent? TryGetEventPayload<TEvent>(this EventSource eventSource)
+        where TEvent : class, IPersistedEvent
+        => eventSource.EventPayload switch
```
IEvangelist, line 30:
```suggestion
        where TEvent : class, IPersistedEvent =>
        eventSource.EventPayload switch
```

nit

IEvangelist (dismissed):

IEvangelist (approved):

PR #223, opened 2022-03-06
Try get async

mumby0168:


src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Read.cs
```diff
@@ -5,15 +5,30 @@
...
 using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.CosmosRepository.Logging;
 
+// ReSharper disable once CheckNamespace
 namespace Microsoft.Azure.CosmosRepository
 {
-    internal sealed partial class DefaultRepository<TItem>
+    internal sealed partial class DefaultRepository<TItem> where TItem : class, IItem
     {
+        public async ValueTask<TItem?> TryGetAsync(string id, string? partitionKeyValue = null, CancellationToken cancellationToken = default)
```
IEvangelist, line 19:
```suggestion
        public async ValueTask<TItem?> TryGetAsync(
            string id,
            string? partitionKeyValue = null,
            CancellationToken cancellationToken = default)

```

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Read.cs
```diff
@@ -5,15 +5,30 @@
...
+        {
+            try
+            {
+                return await GetAsync(id, partitionKeyValue, cancellationToken);
+            }
+            catch (CosmosException e) when (e.StatusCode is HttpStatusCode.NotFound)
+            {
+                _logger.LogItemNotFoundHandled<TItem>(id, partitionKeyValue ?? id, e);
+                return null;
```
IEvangelist, line 28:
```suggestion
                return default;
```

If you use the `default` keyword here instead, you do not need the generic-type `class` constraint.

src/Microsoft.Azure.CosmosRepository/Repositories/InMemoryRepository.cs
```diff
@@ -61,6 +62,18 @@ internal class InMemoryRepository<TItem> : IRepository<TItem> where TItem : clas
         internal TDeserializeTo DeserializeItem<TDeserializeTo>(string jsonItem) =>
             JsonConvert.DeserializeObject<TDeserializeTo>(jsonItem);
 
+        public async ValueTask<TItem?> TryGetAsync(string id, string? partitionKeyValue = null, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                return await GetAsync(id, partitionKeyValue, cancellationToken);
+            }
+            catch (CosmosException e) when (e.StatusCode is HttpStatusCode.NotFound)
```
IEvangelist, line 71: Are we internally (within the delegated `GetAsync`) throwing a `CosmosException`? If not, we'd probably want to alter this a bit, no?

src/Microsoft.Azure.CosmosRepository/Repositories/InMemoryRepository.cs
```diff
@@ -61,6 +62,18 @@ internal class InMemoryRepository<TItem> : IRepository<TItem> where TItem : clas
...
+            {
+                return null;
```
IEvangelist, line 73:
```suggestion
                return default;
```

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.cs
```diff
@@ -1,97 +1,98 @@
...
+using Microsoft.Azure.CosmosRepository.Specification.Evaluator;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Azure.CosmosRepository
+{
+    /// <inheritdoc/>
+    internal sealed partial class DefaultRepository<TItem> : IRepository<TItem>
+        where TItem : class, IItem
```
IEvangelist, line 23: I believe that we should be able to drop the `class` constraint here. Also, we only technically need one implementation of the `partial` to have the generic type constraints. That is why I only had it on this core implementation here (I'm ok with having it on both though, if you prefer that).

IEvangelist (approved):
A few nits, and suggestions.

src/Microsoft.Azure.CosmosRepository/Repositories/InMemoryRepository.cs
mumby0168: We are throwing that in there, the method `NotFound()` throws one.

src/Microsoft.Azure.CosmosRepository/Repositories/InMemoryRepository.cs
mumby0168: This is cool I did not know this!

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.cs
mumby0168: Yeah I did think that, I think I did a bad merge. Updated.

IEvangelist (approved):

PR #227, opened 2022-03-09
Event Sourcing Re Branding, Added some defaults

mumby0168:
I wasn't sure on the names so I did a little renaming.

`IPersistedEvent` -> `IDomainEvent`
`EventSource` -> `EventItem`

I also added some defaults that set the event name and the time stamp so this is something a consumer has to do.

I also added a `Sequence` number to help order the events.

### DIFF
--- a/src/Microsoft.Azure.CosmosEventSourcing/Items/DefaultEventItem.cs
+++ b/src/Microsoft.Azure.CosmosEventSourcing/Items/DefaultEventItem.cs
@@ -35,6 +35,5 @@ public class DefaultEventItem : EventItem
     /// </summary>
     protected DefaultEventItem()
     {
-
     }
 }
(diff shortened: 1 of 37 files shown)


src/Microsoft.Azure.CosmosEventSourcing/Items/DefaultEventItem.cs
```diff
@@ -0,0 +1,40 @@
...
+    [JsonIgnore]
+    public DomainEvent DomainEventPayload =>
+        (DomainEvent) EventPayload;
+
+    /// <summary>
+    /// Creates an <see cref="DefaultEventItem"/>
+    /// </summary>
+    protected DefaultEventItem()
+    {
+
```
IEvangelist, line 38:
```suggestion
```

IEvangelist (dismissed):

IEvangelist (approved):

PR #229, opened 2022-03-09
Atomic event batches

### LINKED ISSUE #224: Save events as atomic batches
When saving a set of events we want all of the events to be saved or none of the events to be saved. This would be supported with a combination of `etags` and the batch support provided by the SDK. 

A possible implementation of this could use a batch operation to create the new events and update an existing special event. This special event can be named something along the lines of `TransactionalEvent` this event can be read and then saved back as part of the batch with a new etag this would mean only one writer of a set of events would win. The other would fail with a pre-condition error as you would expect working with a single item.

This relies on #151

mumby0168:
This implementation aims to use the newly added batch update feature to save all newly added events in a batch, so they either all succeed or none of them do.

The competing consumer's case where two consumers concurrently loaded the current state and apply their changes then both try and write that state change back to the DB is also addressed. 

It is crucial that only one of the consumers in the above case wins. This is optimistic concurrency control, the library already has support for `tags`. I have used this support to always read and update an event called the `AtomicEvent` this event must be read back and _must_ be used when saving any new events.

This in combination with the batches ensures only one consumer in the above example would ever win.

Closes #224


src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Azure.CosmosEventSourcing.Stores;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.CosmosEventSourcing.Events;
+
+/// <summary>
+/// An event that is used to pass between <see cref="IAggregateRoot"/>'s and the <see cref="IEventStore{TEventItem}"/>
+/// <remarks>This is used to provide optimistic concurrency control for batches of <see cref="DomainEvent"/>'s.</remarks>
+/// </summary>
+public record AtomicEvent : DomainEvent
```
mumby0168, line 14: @IEvangelist any ideas on a better name for this? It's basically a marker that needs to be updated with every batch of new events. Struggle with naming lol

tests/Microsoft.Azure.CosmosEventSourcingTests/Aggregates/AggregateRootTests.cs
```diff
@@ -0,0 +1,197 @@
...
+        //Act
+        SampleAggregateRoot root = new(message);
+
+        //Assert
+        root.NewEvents.Should().HaveCount(1);
+        root.Messages.Should().Contain("A");
+    }
+
+    [Fact]
+    public void Ctor_FirstEvent_SetsAtomicEvent()
```
mumby0168, line 82:
```suggestion
    public void Ctor_FirstEvent_CreatesAtomicEvent()
```

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
IEvangelist: I dotn't mind `AtomicEvent`, another name might be `BatchEvent`?

samples/BasicEventSourcingSample/Infrastructure/ShipEventItem.cs
```diff
@@ -9,11 +9,19 @@ namespace BasicEventSourcingSample.Infrastructure;
 
 public class ShipEventItem : DefaultEventItem
 {
-    public ShipEventItem(DomainEvent eventPayload, string shipName)
+
```
IEvangelist, line 12:
```suggestion
```

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
mumby0168: If that makes sense to you happy to keep it as is 😃

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    [JsonConstructor]
+    internal AtomicEvent(Guid id, string eTag)
+    {
+        Id = id;
+        ETag = eTag;
+    }
+
+    internal Guid Id { get; set; }
```
IEvangelist, line 23:
```suggestion
    internal Guid Id { get; init; }
```

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    internal string ETag { get; set; }
```
IEvangelist, line 25:
```suggestion
    internal string ETag { get; init; }
```

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
IEvangelist, line 14:
```suggestion
public record class AtomicEvent : DomainEvent
```

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.cs
```diff
@@ -19,11 +21,25 @@ internal class DefaultEventStore<TEventItem> :
...
-            records,
+        IEnumerable<TEventItem> items,
+        CancellationToken cancellationToken = default)
+    {
+        List<TEventItem> eventItems = items.ToList();
+
+        if (!eventItems.Any())
+        {
+            return;
+        }
```
IEvangelist, line 32:
```suggestion
        if (eventItems is { Count: 0 })
        {
            return;
        }
```

IEvangelist (approved):
A few nits, but otherwise this looks great!

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
mumby0168: Ah not sure it liked this, I think I tried it.

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
IEvangelist: A `record` is implicitly a `class`. With C# 10, you can also now have `record struct`. This is just the more explicit form, saying that you do indeed want a `class` for your `record` type. It's just a style preference, won't change anything here.

https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/record

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
IEvangelist: Ah, are you assigning to this from anything other than the `.ctor` or the object initializer? If not, then this should work.

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
mumby0168: Ahh okay, happy to add this thanks for the clarification.

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
mumby0168: Ah yes, I do update this in the `AggregateRoot`

mumby0168: Screenshot 2022-03-10 at 14 14 25
Ah yes this was it!

mumby0168: Is this just a NuGet package I need to add?

IEvangelist: Ah, ok. I see the problem. This is a project that targets `netstandard2.1`. There is a common `Compat.cs` class that adds the markers for this, see:

https://github.com/IEvangelist/blazorators/blob/main/src/Blazor.SourceGenerators/Compat.cs

```csharp
using System.ComponentModel;

namespace System.Runtime.CompilerServices;

[EditorBrowsable(EditorBrowsableState.Never)]
internal class IsExternalInit { }
```

Also see this: https://stackoverflow.com/a/62656145/2410379

If you add the _Compat.cs_ file, it will compile but more importantly at runtime the actual record `init` bits will function as expected.

src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
mumby0168: Sorted.

PR #246, opened 2022-03-13
Refactored DefaultEventStore into partials & and added ReadAggregateAsync method

mumby0168:
Used partials, introduced read vs write interfaces, added a method that looks for a static Replay method.

Added the concept of mappers to map to and from `AggregateRoot`'s

### DIFF
--- a/src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
+++ b/src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
@@ -7,7 +7,7 @@ using Microsoft.Azure.CosmosRepository;
 namespace Microsoft.Azure.CosmosEventSourcing.Stores;
 
 internal partial class DefaultEventStore<TEventItem> :
-    IEventStore<TEventItem> where TEventItem : EventItem
+    IEventStore<TEventItem> where TEventItem : DefaultEventItem
 {
     private readonly IRepository<TEventItem> _repository;
 
--- a/src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
+++ b/src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
@@ -1,0 +1,16 @@
+// Copyright (c) IEvangelist. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosEventSourcing.Items;
+using Microsoft.Azure.CosmosRepository;
+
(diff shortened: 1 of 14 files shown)


src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Azure.CosmosEventSourcing.Stores;
+
+internal partial class DefaultEventStore<TEventItem> :
+    IEventStore<TEventItem> where TEventItem : DefaultEventItem
+{
+    private readonly IRepository<TEventItem> _repository;
+
+    public DefaultEventStore(IRepository<TEventItem> repository) =>
+        _repository = repository;
```
IEvangelist, line 15: It might make sense to get the `IReadOnlyRepository` and `IWriteOnlyRepository` from DI, and then use each in the respective _*.Read.cs_ and _*.Write.cs_ files.

tests/Microsoft.Azure.CosmosEventSourcingTests/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.Azure.CosmosEventSourcingTests.Extensions;
+
+public static class ObjectExtensions
+{
+    public static void SetPrivatePropertyValue<T>(this object obj, string propName, T val)
```
IEvangelist, line 11: Code smell, why are we doing reflection-based `private` property setting? There are other ways to do this more efficiently and reliably. For example, `internal` and then you could set the visibility to this lib.

IEvangelist (approved):

tests/Microsoft.Azure.CosmosEventSourcingTests/Extensions/ObjectExtensions.cs
mumby0168: Yes i totally agree, going to do some refactoring of the EventItem to make this no an issue, just had to wait while the other PR was merged to avoid conflicts, plan to sort this within the day 👍🏼

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
mumby0168: Yeah I agree I like that idea! I will raise this as another PR along with the work on EventItem

PR #248, opened 2022-03-13
Event item refactor & Acceptance Tests

mumby0168:
Introduced some acceptance tests
1. Tests the persist methods
2. Tests the read methods
3. Tests the change feed processing to build projections
4. Used a single test since the startup time of the library to create containers and DBs etc takes around 20 seconds so running multiple tests just takes longer.

Refactored the `DefaultEventStore<T>` to use the specific interfaces provided by the repo.

Updated docs to reflect changes to `EventItem`. This imp of `EventItem` makes serialization behave better and does not force the consumer to follow any specific ctor implementation.


IEvangelist (approved):

PR #262, opened 2022-03-20
Support for multiple projection builders per host

mumby0168:
The library as per the last version only supported one version of an `IEventItemProjectionBuilder<T>`. If you added more then it would replace the existing implementation.

This PR introduces support for multiple different builders with different responsibilities and importantly using a different processor to build up this projection. It does this through the use of an `IProjectionKey`. This is just a marker type used to aid the DI resolution of the different types.

@IEvangelist thoughts on the `IProjectionKey`? I could not see a cleaner way and thought it was quite clear what the intentions of the processors are this way. I did consider using strings etc but chose to lean on the type system to do a lot of the work.


IEvangelist (approved):

IEvangelist:
@mumby0168 

> Thoughts on the `IProjectionKey`? I could not see a cleaner way and thought it was quite clear what the intentions of the processors are this way. I did consider using strings etc but chose to lean on the type system to do a lot of the work.

It makes sense to me.

PR #269, opened 2022-04-07
Fix ETag Bug with Batch Updates

mumby0168:
We had an issue where the sequence number of events where being duplicated this was because of the Etags and optimistic concurrency not propagating properly between the `AggregateRoot` and the `EventItem` through the `AtomicEvent`'s.


src/Microsoft.Azure.CosmosEventSourcing/Events/AtomicEvent.cs
```diff
@@ -25,5 +25,9 @@ public AtomicEvent(string eventId, string eTag)
         ETag = eTag;
     }
 
+    /// <summary>
+    ///
+    /// </summary>
```
mumby0168, line 30:
```suggestion
```

IEvangelist (approved):

PR #279, opened 2022-05-18
Ability to opt out auto resource creation

mumby0168:
This defaults to the original behaviour of create if not exists. This is great for local dev and small projects.

Larger projects might use infrastructure as code to provision databases and containers. This means the create call is not needed. A small performance benefit can be gained from a read vs a create if not exists.

In addition if you were to use a managed identity to connect to cosmos DB. The create if not exists operation requires more permissions than just read so that can be another benefit.

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository/Services/DefaultCosmosContainerService.cs
+++ b/src/Microsoft.Azure.CosmosRepository/Services/DefaultCosmosContainerService.cs
@@ -46,8 +46,11 @@ namespace Microsoft.Azure.CosmosRepository.Services
                 ItemConfiguration itemConfiguration = _cosmosItemConfigurationProvider.GetItemConfiguration(itemType);
 
                 Database database =
-                    await _cosmosClientProvider.UseClientAsync(
-                        client => client.CreateDatabaseIfNotExistsAsync(_options.DatabaseId)).ConfigureAwait(false);
+                    _options.IsAutoResourceCreationIfNotExistsEnabled
+                        ? await _cosmosClientProvider.UseClientAsync(
+                            client => client.CreateDatabaseIfNotExistsAsync(_options.DatabaseId)).ConfigureAwait(false)
+                        : await _cosmosClientProvider.UseClientAsync(
+                            client => Task.FromResult(client.GetDatabase(_options.DatabaseId))).ConfigureAwait(false);
 
                 ContainerProperties containerProperties = new()
                 {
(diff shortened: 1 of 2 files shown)


robertbennett1998 (approved):
LGTM

src/Microsoft.Azure.CosmosRepository/Services/DefaultCosmosContainerService.cs
```diff
@@ -46,8 +46,11 @@ private async Task<Container> GetContainerAsync(Type itemType, bool forceContain
                 ItemConfiguration itemConfiguration = _cosmosItemConfigurationProvider.GetItemConfiguration(itemType);
 
                 Database database =
-                    await _cosmosClientProvider.UseClientAsync(
-                        client => client.CreateDatabaseIfNotExistsAsync(_options.DatabaseId)).ConfigureAwait(false);
+                    _options.IsAutoResourceCreationIfNotExistsEnabled
+                        ? await _cosmosClientProvider.UseClientAsync(
+                            client => client.CreateDatabaseIfNotExistsAsync(_options.DatabaseId)).ConfigureAwait(false)
+                        : await _cosmosClientProvider.UseClientAsync(
+                            client => Task.FromResult(client.GetDatabase(_options.DatabaseId))).ConfigureAwait(false);
```
IEvangelist, line 53: Can't we get a little closer to where the ternary is needed?

```suggestion
                    await _cosmosClientProvider.UseClientAsync(client =>
                        _options.IsAutoResourceCreationIfNotExistsEnabled
                            ? client.CreateDatabaseIfNotExistsAsync(_options.DatabaseId).ConfigureAwait(false)
                            : Task.FromResult(client.GetDatabase(_options.DatabaseId));
```

src/Microsoft.Azure.CosmosRepository/Services/DefaultCosmosContainerService.cs
```diff
@@ -60,10 +63,13 @@ private async Task<Container> GetContainerAsync(Type itemType, bool forceContain
                 };
 
                 Container container =
-                    await database.CreateContainerIfNotExistsAsync(
-                        containerProperties, itemConfiguration.ThroughputProperties).ConfigureAwait(false);
+                    _options.IsAutoResourceCreationIfNotExistsEnabled
+                        ? await database.CreateContainerIfNotExistsAsync(
+                            containerProperties, itemConfiguration.ThroughputProperties).ConfigureAwait(false)
+                        : await Task.FromResult(database.GetContainer(containerProperties.Id)).ConfigureAwait(false);
```
IEvangelist, line 69:
```suggestion
                    _options.IsAutoResourceCreationIfNotExistsEnabled
                        ? await database.CreateContainerIfNotExistsAsync(
                            containerProperties, itemConfiguration.ThroughputProperties).ConfigureAwait(false)
                        : await Task.FromResult(database.GetContainer(containerProperties.Id));
```

Do we need to call `ConfigureAwait` on the from result calls?

IEvangelist (approved):

PR #290, opened 2022-07-24
Better error handling on non-deserializable event types - favor scoped handlers

mumby0168:
- Support for a more robust deserialization routine using a new `NonDeserializableEvent` to let a consumer detect errors better.
- Support for scoped handlers (ahead of some correlation ID planned work)
- Docs on the new `NonDeserializableEvent` and how it is used, and why it is important to handle

### DIFF
--- a/src/Microsoft.Azure.CosmosEventSourcing/Projections/DefaultDomainEventProjection.cs
+++ b/src/Microsoft.Azure.CosmosEventSourcing/Projections/DefaultDomainEventProjection.cs
@@ -28,7 +28,8 @@ internal class
     {
         string payloadTypeName = eventItem.DomainEvent.GetType().Name;
         Type handlerType = BuildEventProjectionHandlerType(eventItem);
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.CosmosEventSourcing/Projections/DefaultDomainEventProjection.cs
```diff
@@ -40,6 +41,15 @@ public async ValueTask ProjectAsync(TEventItem eventItem, CancellationToken canc
             return;
         }
 
+        if (eventItem.DomainEvent is NonDeserializableEvent nonDeserializableEvent)
+        {
+            _logger.LogError(
+                "The event with name {EventName} could not be deserialized as it was not registered with the custom deserialized payload = {EventPayload}",
```
mumby0168, line 47:
```suggestion
                "The event with name {EventName} could not be deserialized as it was not registered with the custom deserialiser payload = {EventPayload}",
```

IEvangelist (approved):

PR #294, opened 2022-08-02
Upgrade all NuGet bits, and fix issues. And address #287

### LINKED ISSUE #197: IAsyncEnumerable<TItem> Continuation Token Paging Implementation
This library currently has a few paging implementations. None of which yet support streaming. It would be great to add one that returns an `IAsyncEnumerable<TItem>` for scenarios where data would benefit from being streamed. This is a great fit for use with continuation tokens.

A possible API might be something like below:

```csharp
IAsyncEnumerable<TItem> PageAsync(
            Expression<Func<TItem, bool>> predicate = null,
            int? maxResults = null,
            int pageSize = 25,
```

This could potentially allow a caller to either get all the results returned by the `predicate` or specify a limit of how many items they would like back.

> This under the hood would make use of continuation tokens to provide efficient, cost-effective paging.

### LINKED ISSUE #287: Event sourcing Integration Testing with InMemoryRepository fails due to Problems with Dependency Injection
**Describe the bug**
I have a ASP.NET project where I use the Event Sourcing functionality from this SDK. Now I want to use the InMemoryRepository for Integration Testing of my project. But I have identified several problems when I try to do this. But in this bug, I will only care about the two related to Dependency Injection of the InMemoryRepository in the DefaultEventStore constructor.

**Problem 1**
**DefaultEventStore.Def.cs** injects two repositories in the constructor **IBatchRepository<TEventItem>** and **IReadOnlyRepository<TEventItem>**.

Only  **IReadOnlyRepository<>** is registered when calling **builder.Services.AddInMemoryCosmosRepository();**.
This causes problems where **DefaultRepository** is used for writes and **InMemoryRepository** is used for reads in **DefaultEventStore**.
This can easily be solved by adding : (to [src/Microsoft.Azure.CosmosRepository/Extensions/ServiceCollectionExtensions.cs](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/tree/main/src/Microsoft.Azure.CosmosRepository/Extensions/ServiceCollectionExtensions.cs)  AddInMemoryCosmosRepository method )
```
.AddSingleton(typeof(IBatchRepository<>), typeof(InMemoryRepository<>))
```

**Problem 2**
The DI service will now create two different instances of **InMemoryRepository** when creating **DefaultEventStore** after we have applied the fix from Problem 1. This will again give problem that we cannot read what we write in the **DefaultEventStore**. This will directly generate problems related to the AtomicEvent Etag.
The simplest way to fix this is to rewrite the constructor in [src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs) from
``` csharp
public DefaultEventStore(
    IBatchRepository<TEventItem> batchRepository,
    IReadOnlyRepository<TEventItem> readOnlyRepository)
{
    _batchRepository = batchRepository;
    _readOnlyRepository = readOnlyRepository;
}
```
to
``` csharp
public DefaultEventStore(
    IRepository<TEventItem> repository)
{
    _batchRepository = repository;
    _readOnlyRepository = repository;
}
```

**Notes**

With my proposal for fix Problem 2, the fix from Problem 1 will be in relevant in this case, but I think that should be fixed anyway.
Related issue: #266 

**Environment summary**
SDK Version: latest main branch
OS Version: Windows

**Additional context**
Program.cs
``` csharp
builder.Services.AddCosmosEventSourcing(eventSourcingBuilder =>
{
    eventSourcingBuilder.AddCosmosRepository(options =>
    {
options.DatabaseId = "gateway-sample-db";
        options.ContainerBuilder
            .ConfigureEventItemStore<XEventItem>(
                "x-events", containerOptionsBuilder: options => { options.WithServerlessThroughput(); })
            .ConfigureProjectionStore<XReadItem>(
                containerName: "x-projections",
                partitionKey: "/tenantId",
                containerOptionsBuilder: options => { options.WithServerlessThroughput(); });
    });

    eventSourcingBuilder
        .AddEventItemProjection<XEventItem,
            ReadProjectionKey,
            XReadProjection>(
            options =>
            {
                options.ProcessorName = "x-read-projection-builder";

                options.InstanceName =  Environment.MachineName;
            });

    eventSourcingBuilder.AddDomainEventTypes(typeof(Program).Assembly);
});

builder.Services.RemoveCosmosRepositories();

builder.Services.AddInMemoryCosmosRepository();
```

IEvangelist:
- Fixes #287
- Update all NuGet packages and TFMs
- Fix issues related to upgrading
- Add `IAsyncEnumerabe PageAsync` API, fixes #197


src/Microsoft.Azure.CosmosRepository.AspNetCore/Microsoft.Azure.CosmosRepository.AspNetCore.csproj
```diff
@@ -41,12 +41,12 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="3.1.22" />
-        <PackageReference Include="MinVer" Version="2.5.0">
+        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.0" />
```
mumby0168, line 44: Does this mean people using netcore3.1 would not be able to use the library? Not sure it's a huge issue, given that upgrade should have been done by now. But this could also pose a problem potentially for people running on net5?

src/Microsoft.Azure.CosmosRepository/Logging/LoggerMessageDefinitions.cs
```diff
@@ -35,6 +35,13 @@ internal static class LoggerMessageDefinitions
                 LogLevel.Information,
                 EventIds.CosmosQueryConstructed,
                 "Cosmos query constructed for item type {CosmosItemType}: {CosmosQuery}"
+            ); // RanToCompletion
+
+        internal static readonly Action<ILogger, string, string, Exception> RanToCompletion =
```
mumby0168, line 40: For another time, but we ought to look at moving over to the source generators.

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -102,5 +103,44 @@ internal sealed partial class DefaultRepository<TItem>
...
+            if (returnTotal)
+            {
+                total = (await query.CountAsync(cancellationToken))?.Resource;
+            }
+
+            query = maxNumber is int max and > 0 ? query.Take(max) : query;
+
+            _logger.LogQueryConstructed(query);
+
+            IEnumerable<TItem>? items = await GetByQueryAsync(
```
mumby0168, line 133: Would this implementation, using a non-async enumerable approach to load all the records let's say 10 pages into memory, before then starting to stream them back to the client? Would we want to stream the first page back to the caller as soon as it's available to us?

mumby0168 (commented):
Just a few questions for this one, looks great, looking forward to having the async enumerable API.

Might be me, but did this PR mention fixing the in-memory repository bug? I couldn't spot the code for that.

src/Microsoft.Azure.CosmosRepository.AspNetCore/Microsoft.Azure.CosmosRepository.AspNetCore.csproj
IEvangelist: That's a good callout, I'm not sure. I'll have to ask around, but I was under the impression that since that package supports .NET Standard 2.0, it should work.

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
IEvangelist: Yeah, that is a really good callout. I was trying to think of a more appropriate approach. Perhaps, I should refactor this to:

- Get the count first.
- Using the count determine a reasonable page size.
- Page over the data, then for each page yield it back.

That's probably better, right? Or do you have another idea?

IEvangelist: Thoughts on this:
https://github.com/IEvangelist/azure-cosmos-dotnet-repository/pull/294/files#diff-7e1d408ab862dfcc50eebd4168a676306ab6f4c74ef877d6527375f6b7c688dfR129-R147

mumby0168: Yeah that could work, but I am pretty sure running a count query across a large data set can rack up a hefty RU bill, @robertbennett1998 did you have the docs or article we read around this?

I am wondering whether we allow the caller to specify there chunk size, we can assume a reasonable default maybe?

src/Microsoft.Azure.CosmosRepository.AspNetCore/Microsoft.Azure.CosmosRepository.AspNetCore.csproj
mumby0168: I'm sure we had a issue raised about this in the past, I will try and dig it out.

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
IEvangelist: I'm asking one of the PMs on the team, I'm not sure if there is any way to avoid this. Perhaps we just omit the count?

mumby0168: Could the continuation token being null let you know there are no more pages?

src/Microsoft.Azure.CosmosRepository.AspNetCore/Microsoft.Azure.CosmosRepository.AspNetCore.csproj
mumby0168: Did you hear anything back on this @IEvangelist ?

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
mumby0168: Any news on this?

IEvangelist: I was trying to ask this question to @jcocchi. Let's ask her here, and see if this additional context helps to make the question more clear.

src/Microsoft.Azure.CosmosRepository.AspNetCore/Microsoft.Azure.CosmosRepository.AspNetCore.csproj
IEvangelist: Sorry, it's been all quiet on my end. I'll ask again. I'll let you know as soon as I can.

IEvangelist: I confirmed, that this package multi-targets and one of the TFMs is `netstandard2.0`. So this is fine.

PR #307, opened 2022-11-04
Expanding Patch Operations

### LINKED ISSUE #297: Exapand IPatchOperationBuilder to support Increment
> Is this something you are un-sure about or want help with? If so please use the [discussion feature](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/discussions). We can always create an issue after some discussion later if required 😊

I would like to use the IPatchOperationBuilder to increment a property on a document without reading it first.

Expand the IPatchOperationBuilder to support additional patch operations, starting with Increment

Currently I am writing patch operations (outside of replace) manually and would like to use the library to help with that information.

**Additional context**
I looked to see if this was requested before and I did not see anything referencing Increment directly, if this is related to another issue, please let me know and I can close this.

evancjohnson:
This PR expands the patch operations to support Set, Add, Remove, and Increment.
 
@mumby0168 I understand this was assigned to you, I hope I am not stepping on your toes, this is just a draft to make sure I haven't overstepped, I will add tests after I know if this PR is needed.

I also had a question about the _rawPatchOperations is it needed?

Fixes #297

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository/Builders/IPatchOperationBuilder.cs
+++ b/src/Microsoft.Azure.CosmosRepository/Builders/IPatchOperationBuilder.cs
@@ -28,5 +28,73 @@ namespace Microsoft.Azure.CosmosRepository.Builders
         /// <remarks>This currently only supports operations on properties on the root level of a JSON document,
         /// replacing properties on a nested object for example are currently not supported.</remarks>
         IPatchOperationBuilder<TItem> Replace<TValue>(Expression<Func<TItem, TValue>> expression, TValue? value);
+
+        /// <summary>
+        /// Allows a property of an <see cref="IItem"/> to be set with the value provided
+        /// </summary>
+        /// <typeparam name="TValue">The type of the property that is being set.</typeparam>
+        /// <param name="expression">The expression to define which property to operate on.</param>
+        /// <param name="value">The value to set the property defined with.</param>
+        /// <returns>The same instance of <see cref="IPatchOperationBuilder{TItem}"/></returns>
+        IPatchOperationBuilder<TItem> Set<TValue>(Expression<Func<TItem, TValue>> expression, TValue? value);
+
+        /// <summary>
+        /// Allows a property of an <see cref="IItem"/> to be set with the value provided
+        /// </summary>
+        /// <typeparam name="TValue">The type of value to set.</typeparam>
+        /// <param name="path">The json document path</param>
+        /// <param name="value">The value to set.</param>
+        /// <returns>The same instance of <see cref="IPatchOperationBuilder{TItem}"/></returns>
+        IPatchOperationBuilder<TItem> Set<TValue>(string path, TValue? value);
+
+        /// <summary>
+        /// Allows a property of an <see cref="IItem"/> to be added with the value provided
+        /// </summary>
+        /// <typeparam name="TValue">The type of the property that is being added.</typeparam>
+        /// <param name="expression">The expression to define which property to operate on.</param>
+        /// <param name="value">The value to add to the document</param>
+        /// <returns>The same instance of <see cref="IPatchOperationBuilder{TItem}"/></returns>
+        IPatchOperationBuilder<TItem> Add<TValue>(Expression<Func<TItem, TValue>> expression, TValue? value);
+
+        /// <summary>
+        /// Allows a property of an <see cref="IItem"/> to be added with the value provided
+        /// </summary>
+        /// <typeparam name="TValue">The type of the property that is being added.</typeparam>
+        /// <param name="path">Target location reference.</param>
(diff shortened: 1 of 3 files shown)


IEvangelist:
This looks awesome, thank you for this contribution @evancjohnson! As long as @mumby0168 is good with this, let's make it happen. Yes, please add tests. As for the `_rawPatchOperations` I'm not certain, Billy would know.

IEvangelist:
@all-contributors please add @evancjohnson for code and tests. Thank you 💜

evancjohnson:
@mumby0168 @IEvangelist I see where _rawPatchOperations is used, the current implementation of  InMemoryRepository<TItem> expects property info to be provided in the internal patch operation. Is this approach reasonable given that the caller can also provide an explicit path? Using the direct path would allow users to patch a nested object.

Should the InMemoryRepository<TItem> use Newtonsoft.Json in some way (I haven't actually looked up if this possible, I believe I have done something like this in the past) to apply the replace operation to the TItem by path removing the need for _rawPatchOperations and instead leaning on the information provided by the PatchOperation

mumby0168:
Hi @evancjohnson yes, from my distant memory the _rawPatchOperations allowed the in-memory repo to support this feature, it was a challenge and I suspect keeping the support for all the patch operations maybe be difficult ... if there is a way to apply the modifications to JSON from the in memory repo that would be great! Check out the json path support on the `JObject` type that might help in this scenario?

IEvangelist:
> Thanks for the tip @mumby0168. I will do some research on a good approach. I've got a quick questions.
> 
> How rigid is the InMemoryRepository? Is it expected to perform exactly the same way as [CosmosDB](https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update#supported-operations)?

Hi @evancjohnson - it's not super rigid. It was intended to be a helper so that you could use it for unit testing, to help ensure other functionality that might rely on the underlying data store. Does that make sense? We know that there are functional gaps, and we're ok with that for now.

evancjohnson:
@IEvangelist I'll have time to work on this during the weekend, sorry for leaving it open for so long.

MatKumpf:
> @IEvangelist I'll have time to work on this during the weekend, sorry for leaving it open for so long.

Will you be able to finish working on this code?

IEvangelist:
> > @IEvangelist I'll have time to work on this during the weekend, sorry for leaving it open for so long.
> 
> Will you be able to finish working on this code?

Hi @MatKumpf - this was something that @evancjohnson was working on, but hasn't updated. If the op is unable to finish this, I could work on it. Unless it's something that you're interested in?

MatKumpf:
> > > @IEvangelist I'll have time to work on this during the weekend, sorry for leaving it open for so long.
> > 
> > 
> > Will you be able to finish working on this code?
> 
> Hi @MatKumpf - this was something that @evancjohnson was working on, but hasn't updated. If the op is unable to finish this, I could work on it. Unless it's something that you're interested in?

I won't conceal, it would be one of the most important functionalities for me 😄 So far I have to in few places create my custom connection with container and use the rest of patch operations on it.

evancjohnson:
@MatKumpf @IEvangelist Sorry for the delay friends I have just finished a project and am about to graduate school, so I have much more time. I am pretty close to finishing this one up. I do have a question which is more architectural for David and Billy. Should test host support patch operations? I personally do not believe that it should, and those tests should be run against the emulator. 

The reason I suggest that due to the fact patch operations have some not at first noticeable quirks. Check the [troubleshooting](https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update-getting-started?tabs=dotnet#troubleshooting) for an example.

What are the thoughts on just supporting testing of patch operations internally and not exposing those in the test harness provided by the library?

mumby0168:
> @MatKumpf @IEvangelist Sorry for the delay friends I have just finished a project and am about to graduate school, so I have much more time. I am pretty close to finishing this one up. I do have a question which is more architectural for David and Billy. Should test host support patch operations? I personally do not believe that it should, and those tests should be run against the emulator.
> 
> The reason I suggest that due to the fact patch operations have some not at first noticeable quirks. Check the [troubleshooting](https://learn.microsoft.com/en-us/azure/cosmos-db/partial-document-update-getting-started?tabs=dotnet#troubleshooting) for an example.
> 
> What are the thoughts on just supporting testing of patch operations internally and not exposing those in the test harness provided by the library?

This is a decision we've made with some of the other features, as in not to support those features in the in memory implementation I think that is also fine in this case due to the complexity, would you agree @IEvangelist ?

IEvangelist:
Hi again @evancjohnson - any updates on this?

IEvangelist:
We can always reopen this PR. But it's been nearly a year since it was first introduced. And months since the OP has replied.

PR #309, opened 2022-11-08
Bump everything up to .NET 7.

IEvangelist:


mumby0168:
LGTM, does this mean any new version people running on the old .net framework or anything less than net 7.0 won't be able to get any updates?

IEvangelist:
> LGTM, does this mean any new version people running on the old .net framework or anything less than net 7.0 won't be able to get any updates?

We should look into multi-targeting to continue to support the older version folks. Do you have any experience with that?

PR #317, opened 2023-01-18
Improve exception messaging when there is no EventItemPartitionKey attribute explicitly defined

boylec:
This one took my co-worker @mkercheval-tva hours to figure out because the message led him to believe that he was required to have an attribute identifying the partition key even though we already had another Aggregate type working without using that attribute which was confounding to him.

Problem turned out to be that he was calling the overload of PersistAsync (in DefaultEventStore.Write) where you don't explicitly specify the partition key.

Just a minor improvement that may save someone time.


mumby0168 (approved):
Looks a good idea to me, thanks for the contribution, hope your enjoying the library! Any ideas are welcome.

boylec:
> Looks a good idea to me, thanks for the contribution, hope your enjoying the library! Any ideas are welcome.

We certainly are, thank you guys for all that you're doing. We're relying heavily on this library in the wild and it has saved us a ton of time.

PR #362, opened 2023-11-02
Optionally disable sequence numbers using optimistic locking (etags)

mumby0168:
- [x] 1. Introduces a new way to sequence events without optimistic locking
- [x] 2. Adds the ability to use in memory testing for events
- [x] 3. Adds the ability to test change feed projections using an in-memory implementation
- [x] 4. Shows how the in-memory library can be used with functional tests via a sample
- [ ] document non-sequenced option along with the aggregate root settings with the config option
- [ ] document how the in-memory testing works

### DIFF
--- a/docs/event-sourcing/src/event-store/persistence/README.md
+++ b/docs/event-sourcing/src/event-store/persistence/README.md
@@ -39,16 +39,192 @@ public class ShipRepository : IShipRepository
 
 ## Aggregate Roots
 
-1. cover the basic implementation of an aggregate root for applying new events with sequencing on
-1. mention partial classes use for splitting state changes and business logic from the apply methods
-    1. ensure to mention that the apply methods MUST succeed every-time once applied 
+In the context of event sourcing, the Aggregate plays a crucial role in managing state changes. Here are its main responsibilities:
+
+ 1. *Track State Changes*: The Aggregate tracks changes to its state by recording a sequence of events. Each event represents a state change. Instead of storing the current state, the Aggregate stores these events. The current state is derived by replaying these events.
+
+1. *Ensure Consistency*: The Aggregate ensures that any changes to its state are consistent. This means that the state of the Aggregate after an event is always valid according to the business rules of the domain.
+
+1. *Handle Commands*: The Aggregate handles commands, which are requests to perform some action. A command may result in state changes, which are captured as events.
+
+1. *Apply Events*: The Aggregate applies events to change its state. This is done in a deterministic way, meaning given the same sequence of events, the final state of the Aggregate will always be the same.
+
+Remember, the key idea behind an Aggregate in event sourcing is that the state is not directly modified. Instead, changes are captured as events, and the state is derived from these events. This allows for powerful capabilities like temporal querying, event replay, and more.
+
+### Organization
+
+In C#, partial classes allow a single class to be split across multiple files. This can be particularly useful when working with Aggregate Roots in event sourcing for several reasons:
+
+1. *Separation of Concerns*: By using partial classes, you can separate the different responsibilities of the Aggregate Root into different files. For example, you can have one file for handling commands, another for applying events, and another for defining the state. This makes it easier to understand what each part of the Aggregate Root is responsible for.
+
+1. *Readability*: Large classes can be difficult to navigate and understand. By splitting the Aggregate Root into multiple files, you can make the code more readable. Each file can focus on a specific aspect of the Aggregate Root, making it easier to find and understand the relevant code.
+
+1. *Maintainability*: As the complexity of the Aggregate Root grows, having it split into multiple files can make it easier to maintain. Changes to one aspect of the Aggregate Root are less likely to impact other aspects, reducing the risk of introducing bugs.
+
+Remember, the Apply methods in the Aggregate Root are crucial as they are responsible for changing the state of the Aggregate. These methods must always succeed once applied. By separating these methods into their own file, you can highlight their importance and make it easier to ensure they are implemented correctly.
+
+:::tip NOTICE
+It is very important that there is no reason for an `Apply` method to fail. Once an event has been saved into the store it should be considered a fact. If this was to fail in the process of reading back the aggregate and re-applying the events, it would corrupt the state of the aggregate, making it un-readable.
+:::
+
+See below for an example of of splitting the `Ship` aggregate root.
+
+#### Apply
+
+```csharp
+public partial class Ship
+{
+    private Ship() { }
+
+    public static Ship Build(List<DomainEvent> persistedEvents)
+    {
+        Ship ship = new();
+        ship.Apply(persistedEvents);
+        return ship;
+    }
+
+    private void Apply(ShipEvents.ShipCreated shipCreated)
+    {
+        (var name, DateTime commissioned) = shipCreated;
+        Name = name;
+        Commissioned = commissioned;
+    }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.CosmosEventSourcing/Options/CosmosEventSourcingOptions.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) David Pine. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.CosmosEventSourcing.Options;
+
+public class CosmosEventSourcingOptions
```
IEvangelist, line 6: We'll need triple slash for these `public` APIs.

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
```diff
@@ -3,14 +3,18 @@
...
 
 internal partial class DefaultEventStore<TEventItem>(
     IBatchRepository<TEventItem> batchRepository,
     IReadOnlyRepository<TEventItem> readOnlyRepository,
-    IContextService contextService) :
+    IContextService contextService,
+    IOptionsMonitor<CosmosEventSourcingOptions> optionsMonitor) :
     IEventStore<TEventItem> where TEventItem : EventItem
 {
+    private readonly IOptionsMonitor<CosmosEventSourcingOptions> _optionsMonitor = optionsMonitor;
```
IEvangelist, line 19: This field isn't needed, you can access the `optionsMonitor` anywhere within the `partial class`. These are primary constructors.

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Write.cs
```diff
@@ -22,9 +22,12 @@ internal partial class DefaultEventStore<TEventItem>
             return;
         }
 
-        if (eventItems.Count(x => x.EventName is nameof(AtomicEvent)) is not 1)
+        if(_optionsMonitor.CurrentValue.IsSequenceNumberingDisabled is false)
```
IEvangelist, line 25:
```suggestion
        if (optionsMonitor.CurrentValue.IsSequenceNumberingDisabled is false)
```

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Write.cs
```diff
@@ -90,10 +93,13 @@ internal partial class DefaultEventStore<TEventItem>
                     partitionKey) as TEventItem)
             .ToList();
 
-        events.Add(Activator.CreateInstance(
-            typeof(TEventItem),
-            aggregateRoot.AtomicEvent,
-            partitionKey) as TEventItem);
+        if(_optionsMonitor.CurrentValue.IsSequenceNumberingDisabled is false)
```
IEvangelist, line 96:
```suggestion
        if (optionsMonitor.CurrentValue.IsSequenceNumberingDisabled is false)
```

tests/Microsoft.Azure.CosmosEventSourcingTests/Stores/EventStoreTests.ReadAggregate.cs
```diff
@@ -63,6 +63,32 @@ private TestAggregate()
...
+    private class TestAggregateNoSequencing : AggregateRoot
+    {
+        public int ReplayedEvents { get; private set; }
+
+        protected override void Apply(DomainEvent domainEvent)
+        {
+            switch (domainEvent)
+            {
+                case ReplayableEvent:
+                    ReplayedEvents++;
```
IEvangelist, line 75: Any race condition concerns with multiple threads calling apply at the same time, and reads of the `ReplayedEvents`?

src/Microsoft.Azure.CosmosEventSourcing/Stores/DefaultEventStore.Def.cs
mumby0168: ahhh that makes sense now, I'll update this cheers!

tests/Microsoft.Azure.CosmosEventSourcingTests/Stores/EventStoreTests.ReadAggregate.cs
mumby0168: Yes there will be, kind of a long story to this change, I plan to update the docs alongside this as well. But in short, the sequence number and the strong consistency provided by the atomic event is great in a single region scenario.

However, when we have been assessing the ability to enable multi region writes where traffic would be load balanced across two regions both writing to there local cosmos db data store, we could not guarantee that this sequence number remains in sync and in some use cases for event sourcing it doesn't really matter. The plan here is to handle simultaneous writes, as long as that event is valid at the point in time we are going to let it go ahead.

This diagram is covering our use case, but I am still fleshing out all the details of whether this is 100% suitable via some POCs.

Screenshot 2023-11-02 at 18 05 29

src/Microsoft.Azure.CosmosEventSourcing/Extensions/AggregateRootExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        var partitionKeyProperties = aggregate
+            .GetType()
+            .GetProperties()
+            .Where(x
+                => x.GetCustomAttributes()
+                    .Any(y =>
+                        y is EventItemPartitionKeyAttribute))
+            .ToList();
+
+        switch (partitionKeyProperties.Count)
```
IEvangelist, line 54: Prefer switch expressions.

src/Microsoft.Azure.CosmosEventSourcing/Extensions/EventItemExtensions.cs
```diff
@@ -36,4 +39,41 @@ eventItem.DomainEvent switch
             null => null,
             _ => throw new ArgumentOutOfRangeException()
         };
+
+    public static TAggregateRoot Replay<TAggregateRoot, TEventItem>(this IEnumerable<TEventItem> events)
+    where TEventItem : EventItem
```
IEvangelist, line 44:
```suggestion
        where TEventItem : EventItem
```

src/Microsoft.Azure.CosmosEventSourcing/Extensions/EventItemExtensions.cs
```diff
@@ -36,4 +39,41 @@ eventItem.DomainEvent switch
...
+
+        return method is null
+            ? throw new ReplayMethodNotDefinedException(type)
+            : (TAggregateRoot)method.Invoke(null, new object[] { payloads })!;
+    }
+
+    public static IEnumerable<TEventItem> SetCorrelationId<TEventItem>(
+        this IEnumerable<TEventItem> eventItems,
+        IContextService contextService)
+    where TEventItem : EventItem
```
IEvangelist, line 61:
```suggestion
        where TEventItem : EventItem
```

tests/Microsoft.Azure.CosmosEventSourcingTests/Stores/EventStoreTests.ReadAggregate.cs
IEvangelist: I'm good with it if you are, let's :shipit:

PR #370, opened 2023-11-17
CPM and .NET 8.0

IEvangelist:


samples/Microsoft.Azure.CosmosEventSourcing/BasicEventSourcingSample/Program.cs
```diff
@@ -13,13 +13,12 @@
...
 using Microsoft.Azure.CosmosRepository.AspNetCore.Extensions;
-using Microsoft.Azure.CosmosRepository.Extensions;
 
 WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 IServiceCollection services = builder.Services;
 
-
-services.AddHealthChecks().AddCosmosDb(builder.Configuration.GetCosmosRepositoryConnectionString()!);
+// TODO: Figure this out...
+// services.AddHealthChecks().AddCosmosDb(builder.Configuration.GetCosmosRepositoryConnectionString()!);
```
mumby0168, line 21: Has this broken?

Directory.Packages.props
```diff
@@ -0,0 +1,93 @@
...
+    <PackageVersion Include="MartinCostello.Logging.XUnit" Version="0.3.0" />
+    <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.36.0" />
+    <PackageVersion Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
+    <PackageVersion Include="MinVer" Version="5.0.0-beta.1" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.3" />
+    <PackageVersion Include="xunit" Version="2.6.1" />
+    <PackageVersion Include="Scrutor" Version="4.2.2" />
+  </ItemGroup>
+  <!-- .NET 7.0. bits -->
```
mumby0168, line 20: I think I spotted an issue the other day when using the package in azure functions, do you think it's worth adding a .net 6 target and downgrading some of the MS packages when we are targeting net6.0?

mumby0168 (approved):

samples/Microsoft.Azure.CosmosEventSourcing/BasicEventSourcingSample/Program.cs
IEvangelist: Yeah, with the latest version the `AddHealthChecks` method doesn't exist, and I didn't have time to look it up. It was in a sample app so I left it for now, but happy to hold this for a bit.

Directory.Packages.props
IEvangelist: .NET 6 is supported for less than a year remaining, https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle. I personally don't think it's worth doing... especially since we have support for `netstandard2.0` which allows us to use this all over.

samples/Microsoft.Azure.CosmosEventSourcing/BasicEventSourcingSample/Program.cs
IEvangelist: Fixed it...

mumby0168 (approved):
LGTM

PR #399, opened 2024-01-29
add await to async test asserts

stormbringer766:
add missing await to async test asserts causing build failure.


IEvangelist:
This is awesome, thank you for this contribution @stormbringer766 - any chance we could get all the tests to pass?

IEvangelist:
@all-contributors, please add @stormbringer766 for tests.

IEvangelist:
@all-contributors please add @stormbringer766 for tests.

stormbringer766:
@IEvangelist not sure this is the way you want this fixed. There were a couple of issues.
1. The client is `Lazy` so wasn't actually being created in the test so wasn't being disposed.
2. The actual exception is a `CosmosObjectDisposedException` which is internal so can't be asserted.

Not sure what is up with `CodeQL`

```
Error: Code Scanning could not process the submitted SARIF file:
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

IEvangelist (approved):

PR #400, opened 2024-02-02
fix: default spec does not return all results and defaults to 25

robertbennett1998:


philip-reed:
@robennett2 Can you provide an update on this PR?
I think im encountering the issue this would resolve.

PR #418, opened 2024-03-07
feat: Extend HealthChecks to allow AzureCosmosDbHealthCheckOptions t…

### LINKED ISSUE #417: HealthCheck Defaults Cannot be overridden

#377 Introduced support for health checks, however the default behavior of `AspNetCore.HealthChecks.CosmosDb` is to only check that the Account can be read. #377 Does not allow any way to invoke this behavior.

image
[https://www.nuget.org/packages/AspNetCore.HealthChecks.CosmosDb/](https://www.nuget.org/packages/AspNetCore.HealthChecks.CosmosDb/)

Provide a way to configure the health checks to  use the default configuration for the `AspNetCore.HealthChecks.CosmosDb` package.  The simplest solution would be to adopt the same defaults, but this would be a breaking change for anyone already using this feature.  Alternately, some override can be provided to keep the default behavior of `AspNetCore.HealthChecks.CosmosDb` rather than scanning assemblies for the `IItem` container names.

**Additional context**

This use case has come about because heath checks return unhealthy when `IsAutoResourceCreationIfNotExistsEnabled: true` and the app has not created the containers yet.  The health check call does not create the containers, and nor should it.

philip-reed:
Closes #417 

Adds an overload to the `HealthChecksBuilderExtensions` to allow `AzureCosmosDbHealthCheckOptions` to be provided and bypass assembly scanning to look for `IItem` container names.

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository.AspNetCore/Extensions/HealthChecksBuilderExtensions.cs
+++ b/src/Microsoft.Azure.CosmosRepository.AspNetCore/Extensions/HealthChecksBuilderExtensions.cs
@@ -55,4 +55,32 @@ public static class HealthChecksBuilderExtensions
 
         return builder;
     }
+
+    /// <summary>
+    /// Add a health check for Azure Cosmos DB by registering <see cref="AzureCosmosDbHealthCheck"/> for given <paramref name="builder"/>.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHealthChecksBuilder"/> to add <see cref="HealthCheckRegistration"/> to.</param>
+    /// <param name="healthCheckName">The health check name. Optional. If <c>null</c> the name 'azure_cosmosdb' will be used.</param>
+    /// <param name="failureStatus">
+    /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <c>null</c> then
+    /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
+    /// </param>
+    /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
+    /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.CosmosRepository.AspNetCore/Extensions/HealthChecksBuilderExtensions.cs
```diff
@@ -55,4 +55,32 @@ public static class HealthChecksBuilderExtensions
 
         return builder;
     }
+
+    /// <summary>
+    /// Add a health check for Azure Cosmos DB by registering <see cref="AzureCosmosDbHealthCheck"/> for given <paramref name="builder"/>.
```
mumby0168, line 60: I think this looks good, just need to edit the two summary XML tags so it's clear, that one automatically scans for all item types and the other allows you to manually specify

mumby0168 (changes requested):

IEvangelist (approved):

src/Microsoft.Azure.CosmosRepository.AspNetCore/Extensions/HealthChecksBuilderExtensions.cs
philip-reed: Sorted 👍

philip-reed:
@IEvangelist On it. 👍

philip-reed:
@IEvangelist Looks like there's a failed workflow run on main after this was merged, something to do with the docs site.  
Is that something you can help with?

https://github.com/IEvangelist/azure-cosmos-dotnet-repository/actions/runs/8189444429

IEvangelist:
I fixed it, thanks though @philip-reed - your changes will be part of the `8.1.4` NuGet, should be live in about 15 mins: https://github.com/IEvangelist/azure-cosmos-dotnet-repository/releases/tag/8.1.4

Thanks again!

Issue #432, opened 2024-04-19
Add support for QueryRequestOption property ResponseContinuationTokenLimitInKb


victormarante:
**Is your feature request related to a problem? Please describe.**
Continuation tokens from Cosmos V3 SDK sometimes become too large for servers/browser agents too handle (either resulting in errors when receiving such a large item, or getting 414 URI too long). This issue happens when continuation token is part of route, but will also be an issue since it can be too large even for headers.

**Describe the solution you'd like**
QueryRequestOptions from the Cosmos SDK supports setting a continuation token limit. This would be helpful if it was exposed from this NuGet so we can control the size of the token in cases there it becomes to large

IEvangelist:
Hi there @victormarante - I like this proposal. Would you be interested in helping out, by providing a PR? I'm imagining that we could add an overload that exposes this to the various APIs where it's needed.

victormarante:
Hi! I can surely do so. I will look into your contribution guidelines and create a PR once I am done.

victormarante:
Noticed you don't have a contribution guideline, but I will do my best. A question on what direction you are thinking:

1. As you suggested, go through all the API's and add an overload for the API's that _may_ return a ContinuationToken, or
2. Allow users to send in either `QueryRequestOptions` or sub-set of the options to _all_ API's that utilizes it (which seems to be `DefaultRepository.Paging.cs` and `DefaultRepository.Specs.cs`

Not sure which approach you fancy, but the easier on (point 1), might be what I begin with and we'll go from there.

IEvangelist:
Yeah, I think the overload approach is the most straightforward (option 1). That seems similar to option 2, no? Wouldn't the overload accept a `QueryRequestOptions`?

IEvangelist:
@all-contributors please add @victormarante for issue and code

PR #437, opened 2024-05-06
Feature/support continuation token limit property

victormarante:
This PR will add support for the [Cosmos SDK API QueryOptionsClass](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.queryrequestoptions?view=azure-dotnet) which enables consumers of these APIs to provide their own settings for the `CosmosClient`.

This PR aims to solve: #432

@IEvangelist I have made an initial PR here (no tests or so yet) in order to make sure I am on the right track before spending more time on this. I have two questions that will need discussion:

- It might be confusing for users to provide both the query options _AND_ page size, as they might enter the same information twice (if they provided MaxItemCount in query options to begin with). I do not know what your thought is about this.
- Are you ok with the method overloading (i.e. that I simply call the new method from the existing API), with same logic + a default instance of `QueryOptionsClass`

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
+++ b/src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
@@ -3,6 +3,10 @@
 
 
 // ReSharper disable once CheckNamespace
+using System.Threading;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.Options;
+
 namespace Microsoft.Azure.CosmosRepository;
 
 internal sealed partial class DefaultRepository<TItem>
@@ -14,18 +18,35 @@ internal sealed partial class DefaultRepository<TItem>
         string? continuationToken = null,
         bool returnTotal = false,
         CancellationToken cancellationToken = default)
+        => await PageAsync(
+            new QueryRequestOptions() { MaxItemCount = pageSize },
+            predicate,
+            pageSize,
+            continuationToken,
+            returnTotal,
+            cancellationToken);
+
+    /// <inheritdoc/>
+    public async ValueTask<IPage<TItem>> PageAsync(
+        QueryRequestOptions requestOptions,
+        Expression<Func<TItem, bool>>? predicate = null,
+        int pageSize = 25,
+        string? continuationToken = null,
+        bool returnTotal = false,
+        CancellationToken cancellationToken = default)
     {
         Container container = await containerProvider.GetContainerAsync()
             .ConfigureAwait(false);
 
-        QueryRequestOptions options = new()
+        // make sure that if the user hasn't said the value already we take it from the pageSize parameter
+        if (requestOptions.MaxItemCount is null)
         {
-            MaxItemCount = pageSize
-        };
+            requestOptions.MaxItemCount = pageSize;
+        }
 
         IQueryable<TItem> query = container
             .GetItemLinqQueryable<TItem>(
-                requestOptions: options,
+                requestOptions: requestOptions,
                 continuationToken: continuationToken,
                 linqSerializerOptions: optionsMonitor.CurrentValue.SerializationOptions)
             .Where(repositoryExpressionProvider.Build(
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -3,6 +3,10 @@
 
 
 // ReSharper disable once CheckNamespace
+using System.Threading;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.Options;
```
IEvangelist, line 8: These should be moved to the [_GlobalUsings.cs_ file](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/main/src/Microsoft.Azure.CosmosRepository/GlobalUsings.cs).

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -13,19 +17,36 @@ internal sealed partial class DefaultRepository<TItem>
         int pageSize = 25,
         string? continuationToken = null,
         bool returnTotal = false,
-        CancellationToken cancellationToken = default)
-    {
+        CancellationToken cancellationToken = default)
+        => await PageAsync(
```
IEvangelist, line 21: Minor style nit. Prefer the goes to operator to be on the previous line. Also, if refactoring this entry point, to be an expression-bodied member it's likely we could remove `async` and `await` here to remove an extra state machine blip.

```suggestion
        CancellationToken cancellationToken = default) =>
            await PageAsync(
```

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -13,19 +17,36 @@ internal sealed partial class DefaultRepository<TItem>
...
+            new QueryRequestOptions() { MaxItemCount = pageSize },
```
IEvangelist, line 22:
```suggestion
            new QueryRequestOptions { MaxItemCount = pageSize },
```

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -61,13 +82,30 @@ internal sealed partial class DefaultRepository<TItem>
         int pageSize = 25,
         bool returnTotal = false,
         CancellationToken cancellationToken = default)
-    {
+    => await PageAsync(
+            requestOptions: new QueryRequestOptions() { MaxItemCount = pageSize },
```
IEvangelist, line 86:
```suggestion
            requestOptions: new QueryRequestOptions { MaxItemCount = pageSize },
```

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Specs.cs
```diff
@@ -11,22 +11,27 @@ internal sealed partial class DefaultRepository<TItem>
     public async ValueTask<TResult> QueryAsync<TResult>(
         ISpecification<TItem, TResult> specification,
         CancellationToken cancellationToken = default)
-        where TResult : IQueryResult<TItem>
-    {
+        where TResult : IQueryResult<TItem>
+        => await QueryAsync(specification, new QueryRequestOptions(), cancellationToken);
```
IEvangelist, line 15:
```suggestion
        where TResult : IQueryResult<TItem> =>
            await QueryAsync(specification, new QueryRequestOptions(), cancellationToken);
```

IEvangelist (commented):
Looks okay thus far... Looks like some of these are new APIs and one is modified (`PageAsync`). This is a breaking change, it's probably better to add all new APIs.

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
victormarante: - Refactored placement of "goes to"-operator
- Modified passthrough methods to not include async/await

IEvangelist:
@all-contributors please add @victormarante for code

victormarante:
Hi, great, thanks for reviewing this far. I would like to add some tests but unsure what exactly you prefer in terms of tests?

IEvangelist:
> Hi, great, thanks for reviewing this far. I would like to add some tests but unsure what exactly you prefer in terms of tests?

There are a lot of tests that already exist, rely on those for inspiration for new tests.

src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Paging.cs
```diff
@@ -3,29 +3,47 @@
...
-            .ConfigureAwait(false);
-
-        QueryRequestOptions options = new()
-        {
-            MaxItemCount = pageSize
-        };
+            .ConfigureAwait(false);
+
+        // make sure that if the user hasn't said the value already we take it from the pageSize parameter
+        if (requestOptions.MaxItemCount is null)
```
mateuszkumpf, line 39: You must expect that null could be passed. Initialize requestOption if the value is null.

mateuszkumpf: Or throw exception

IEvangelist: But the `requestOptions` isn't nullable, are you implying that we should protect against it anyway?

mateuszkumpf: By default, I don't trust users 😂 But if for you it's OK then for me too 😄

PR #451, opened 2024-06-27
Resolve #425

### LINKED ISSUE #425: Minimum Autoscale Throughput should be as low as 1000 RU/s
**Describe the bug**
As specified in the  [documentation](https://learn.microsoft.com/en-us/azure/cosmos-db/autoscale-faq#what-s-the-entry-point-ru-s-for-autoscale-) the minimum autoscale throughput can be as low as 1000 RU/s from April 2022. The current implementation is using the old minimum value 4000 RU/s.

**To Reproduce**
Try to create a new container using `containerOptionsBuilder.WithAutoscaleThroughput()` using a value `1000 < maxAutoScaleThroughput < 4000 `

**Expected behavior**
The container should be created successfully without any error being thrown.

**Actual behavior**
The following error is thrown `Autoscale throughput must be between 4,000 and 1,000,000 RUs.`

philip-reed:
Resolves #425

### DIFF
--- a/tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
+++ b/tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
@@ -25,7 +25,7 @@ public class DefaultCosmosThroughputProviderTests : WithRepositoryOptions
         _repositoryOptions.ContainerBuilder.Configure<TestItemWithEtag>(builder => builder.WithAutoscaleThroughput());
         ThroughputProperties? throughputProperties = _provider.GetThroughputProperties<TestItemWithEtag>();
 
-        Assert.Equal(4000, throughputProperties!.AutoscaleMaxThroughput);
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

tests/Microsoft.Azure.CosmosRepositoryTests/Providers/DefaultCosmosThroughputProviderTests.cs
```diff
@@ -25,7 +25,7 @@ public void GetThroughputPropertiesItemAutoscaleThroughputProperties()
         _repositoryOptions.ContainerBuilder.Configure<TestItemWithEtag>(builder => builder.WithAutoscaleThroughput());
         ThroughputProperties? throughputProperties = _provider.GetThroughputProperties<TestItemWithEtag>();
 
-        Assert.Equal(4000, throughputProperties!.AutoscaleMaxThroughput);
+        Assert.Equal(1000, throughputProperties!.AutoscaleMaxThroughput);
```
iamteamstar: i did not understand why

iamteamstar (commented):
I think instead of modifying the test to test different scenarios, you should create a new inlinedata under it.

PR #454, opened 2024-06-28
A bit of cleanup and fix a bug where UseStrictTyping didn't flow through specs

### LINKED ISSUE #452: UseStrictTypeChecking ContainerOption is ignored when using DefaultRepository.Specs
Configuring a Repository ContainerOptions with `.WithoutStrictTypeChecking()` STILL appends type checking to specification queries

**Reproducing The Issue**
```
builder.Services.AddCosmosRepository(options => 
{
    options.ContainerBuilder.Configure<ExampleModel>(containerOptions => containerOptions
        .WithoutStrictTypeChecking()
        .WithContainer("Example")
        .WithPartitionKey("/id"));
};
```
```
public class ExampleSpecification : DefaultSpecification<ExampleModel>
{
    public ExampleSpecification() =>
        Query
            .Where(q => q.Category == "Red");
}
```
Main.cs
```
var queryResult = await _cosmosRepository.QueryAsync(new ExampleSpecification());
```

**Console Logged Expected Query**
```
"SELECT TOP 25 VALUE root FROM root WHERE (root[\"category\"] = \"Red\")"
```
**Console Logged Actual Query**
```
"SELECT TOP 25 VALUE root FROM root WHERE (((NOT IS_DEFINED(root[\"type\"])) OR (root[\"type\"] = \"ExampleModel\")) AND (root[\"category\"] = \"Red\"))"
```
In my case I need to retrieve all "Red" objects in the given collection as "ExampleModel"s, but all of my models are stored as type  "OtherModel". The _actual query_ requires "type" exactly match "ExampleModel" or is not defined. So when executed, the _actual query_ does not retrieve anything.

**Environment summary**
SDK Version: 8.1.5
OS Version: Windows

**Additional context**
This is the case since [DefaultRepository.Specs](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/d8d6f5287565a4a37bc9688a63f347d4efbfd950/src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Specs.cs) uses [DefaultRepositoryExpressionProvider](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/d8d6f5287565a4a37bc9688a63f347d4efbfd950/src/Microsoft.Azure.CosmosRepository/Providers/DefaultRepositoryExpressionProvider.cs).Default() which does not check UseStrictTypeChecking like Build() does for other DefaultRepository Operations.

To test, in [DefaultRepository.Specs](https://github.com/IEvangelist/azure-cosmos-dotnet-repository/blob/d8d6f5287565a4a37bc9688a63f347d4efbfd950/src/Microsoft.Azure.CosmosRepository/Repositories/DefaultRepository.Specs.cs) I replaced `repositoryExpressionProvider.Default<TItem>()` with `repositoryExpressionProvider.Build<TItem>(_ => true)` and produced the _expected query_. The popper way to fix this would probably be fixing Default() in the ExpressionProvider to work more similarly to Build().

IEvangelist:
Hey @guydigital, please have a look.

Fixes #452, replaces PR #453

### DIFF
--- a/samples/Microsoft.Azure.CosmosRepository/BugSandbox/Program.cs
+++ b/samples/Microsoft.Azure.CosmosRepository/BugSandbox/Program.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) David Pine. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.CosmosRepository;
+using Microsoft.Azure.CosmosRepository.Specification;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+// https://learn.microsoft.com/azure/cosmos-db/how-to-develop-emulator?tabs=docker-linux%2Ccsharp&pivots=api-nosql
+const string EmulatorConnectionString = """
+    AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==;
+    """;
+
+var builder = Host.CreateApplicationBuilder(args);
+
+ServiceProvider provider = builder.Services.AddCosmosRepository(options =>
+    {
+        options.CosmosConnectionString = EmulatorConnectionString;
+        options.DatabaseId = "bug-sandbox";
+        options.ContainerBuilder.Configure<ExampleModel>(
+            static containerOptions => containerOptions
+                .WithoutStrictTypeChecking()
+                .WithContainer("Example")
+                .WithPartitionKey("/id"));
+    })
+    .BuildServiceProvider();
+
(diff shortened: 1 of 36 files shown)


src/Microsoft.Azure.CosmosRepository/Providers/DefaultRepositoryExpressionProvider.cs
GuyDigital, line 1: Thank You, This cleanly resolves #452

samples/Microsoft.Azure.CosmosRepository/BugSandbox/Program.cs
GuyDigital, line 1: Forgive me. I am still learning how to test this myself.
Since this is a sample, we can prove that the changes are working using the logged query.
If I wanted to turn this into a Test, all I would need to do is make a second repository with the same container and use a different model for seeding, right?

IEvangelist: I manually verified it, and watched the logs correctly produce the expected result. I should have written a test, but this was easier - and I was short on time.

PR #516, opened 2025-03-28
Added check for URI when verifying if emulator connection string

MattHartz:
When using AddCosmosRepositoryClient, options and clientOptions successfully process when using a connection string. But when a named connection string that represents the "AccountEndpoint", options' action successfully processes it as an AccountEndpoint that will use DefaultCredential but  IsEmulatorConnectionString throws an exception when creating the DbConnectionStringBuilder


MattHartz:
Small issue I noticed when playing around test containers. Works just great when using a locally running emulator, but when I specify a cloud cosmos, via just the URI, it would throw exceptions. I found that the culprit it the string extensions method.

PR #528, opened 2026-01-22
Add Eager DB initialisation, and sample app to prove functionality

philip-reed:
Addresses #527 

Adds a new `EagerlyInitializeCosmosContainersAsync` extension which creates containers **at application startup** before any requests are handled.

This allows developers using this library in CI pipelines to wait for DB readiness before handling any requests (specifically health checks).

- Added new extension method for eager DB initialisation
- Added sample app to prove functionality
- Updated docs to explain use cases

> Note, I skipped units test as this is orchestration code, and I wasn't sure if they would add much value.

### DIFF
--- a/docs/cosmos-repo/content/6-miscellaneous/container-initialization/_index.md
+++ b/docs/cosmos-repo/content/6-miscellaneous/container-initialization/_index.md
@@ -1,0 +1,227 @@
+---
+title: "Container Initialization"
+weight: 4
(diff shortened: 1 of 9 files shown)


next-phil:
@IEvangelist Any chance you can give this a look over when you get the time please 🙏

IEvangelist (approved):
LGTM

PR #540, opened 2026-04-19
Add cross-type transactional batch support

### LINKED ISSUE #151: Support for TransactionalBatch
The repositories currently don't offer support for transactions (using TransactionalBatch), for use cases where a bunch of CRUD actions have to all succeed/fail together. If transactions are required, the only solution now is to implement a custom Repository, which kind of defeats the purpose of using this library.

It would be cool to be able to do something along the lines of:
`await _repository.Batch(partitionKey).Create(item1).Replace(item2.Id, item2).ExecuteAsync();`

cjbanna:
Closes #151.

## Summary
- add `IBatchBuilder` and `IBatchRepository<T>.Batch(string)` to support fluent transactional batch operations across item types that share a container
- preserve the existing single-type batch APIs while introducing a non-generic `BatchOperationException` base that keeps `BatchOperationException<TItem>` catch-compatible
- add in-memory batch support, wire up `IBatchRepository<>` for in-memory DI, and cover the new behavior with unit and acceptance tests

Example usage:
```csharp
await _repository
    .Batch(partitionKey)
    .CreateItem(product)
    .CreateItem(rating)
    .DeleteItem<LegacyRating>(legacyId)
    .ExecuteAsync();
```

I also considered a separate batch factory/service that would not hang off `IRepository<TItem>`, but keeping `Batch(string)` on the existing repository felt more ergonomic because the seed repository already conveys the container scope and the semantics here are still "a transactional batch over one container."

## Testing
- `dotnet test tests/Microsoft.Azure.CosmosRepositoryTests/Microsoft.Azure.CosmosRepositoryTests.csproj`

## Compatibility
- This is a non-breaking feature addition.
- Existing `CreateAsBatchAsync`, `UpdateAsBatchAsync`, and `DeleteAsBatchAsync` APIs remain unchanged.
- Existing `catch (BatchOperationException<TItem>)` handlers continue to work.

### DIFF
--- a/src/Microsoft.Azure.CosmosRepository/Batching/DefaultBatchBuilder.cs
+++ b/src/Microsoft.Azure.CosmosRepository/Batching/DefaultBatchBuilder.cs
@@ -1,0 +1,179 @@
+namespace Microsoft.Azure.CosmosRepository;
+
+internal sealed class DefaultBatchBuilder(
+    string partitionKey,
+    Type seedType,
+    ICosmosContainerService containerService,
+    ICosmosItemConfigurationProvider configProvider) : IBatchBuilder
+{
+    private readonly string _partitionKey = partitionKey;
+    private readonly string _seedContainerName = configProvider.GetItemConfiguration(seedType).ContainerName;
+    private readonly List<PendingOp> _operations = [];
+    private readonly HashSet<Type> _seenTypes = [seedType];
+
+    public IBatchBuilder CreateItem<TItem>(TItem item) where TItem : IItem =>
+        Add(OpKind.Create, item);
+
+    public IBatchBuilder ReplaceItem<TItem>(TItem item) where TItem : IItem =>
+        Add(OpKind.Replace, item);
+
+    public IBatchBuilder UpsertItem<TItem>(TItem item) where TItem : IItem =>
+        Add(OpKind.Upsert, item);
+
+    public IBatchBuilder DeleteItem<TItem>(TItem item) where TItem : IItem =>
+        Add(OpKind.Delete, item);
+
+    public IBatchBuilder DeleteItem<TItem>(string id) where TItem : IItem
+    {
+        if (id is null)
+        {
+            throw new ArgumentNullException(nameof(id));
+        }
+
+        ValidateContainer(typeof(TItem));
+        _operations.Add(new(OpKind.Delete, typeof(TItem), null, id, null));
+
+        return this;
+    }
+
+    public async ValueTask ExecuteAsync(CancellationToken cancellationToken = default)
+    {
+        if (_operations.Count == 0)
+        {
+            throw new ArgumentException(
+                "Unable to perform batch operation with no items",
+                nameof(_operations));
+        }
+
+        Container container = await containerService.GetContainerAsync(_seenTypes.ToList())
+            .ConfigureAwait(false);
+
+        TransactionalBatch batch = container.CreateTransactionalBatch(new PartitionKey(_partitionKey));
+
+        foreach (PendingOp operation in _operations)
+        {
+            switch (operation.Kind)
+            {
+                case OpKind.Create:
+                    batch.CreateItem(operation.Item);
+                    break;
+                case OpKind.Replace:
+                    batch.ReplaceItem(operation.Id, operation.Item, CreateRequestOptions(operation.Etag));
+                    break;
+                case OpKind.Upsert:
+                    batch.UpsertItem(operation.Item, CreateRequestOptions(operation.Etag));
+                    break;
+                case OpKind.Delete:
+                    batch.DeleteItem(operation.Id);
+                    break;
+                default:
+                    throw new InvalidOperationException($"Unknown batch operation kind: {operation.Kind}");
+            }
+        }
+
+        TransactionalBatchResponse response = await batch.ExecuteAsync(cancellationToken)
+            .ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new BatchOperationException(response);
+        }
+
+        response.Dispose();
+    }
+
+    private IBatchBuilder Add<TItem>(OpKind kind, TItem item) where TItem : IItem
+    {
+        if (item is null)
+        {
+            throw new ArgumentNullException(nameof(item));
+        }
+
+        ValidateContainer(typeof(TItem));
+        ValidatePartitionKey(item);
+
+        _operations.Add(new(
+            kind,
+            typeof(TItem),
+            item,
+            item.Id,
+            (kind is OpKind.Replace or OpKind.Upsert) && item is IItemWithEtag itemWithEtag
+                ? itemWithEtag.Etag
+                : null));
+
+        return this;
+    }
+
+    private void ValidateContainer(Type itemType)
+    {
+        if (!_seenTypes.Add(itemType))
+        {
+            return;
+        }
+
+        string containerName = configProvider.GetItemConfiguration(itemType).ContainerName;
+
+        if (!string.Equals(containerName, _seedContainerName, StringComparison.Ordinal))
+        {
+            throw new InvalidOperationException(
+                $"The item type {itemType.Name} is not configured to use the same container as {seedType.Name}.");
+        }
+    }
+
+    private void ValidatePartitionKey(IItem item)
+    {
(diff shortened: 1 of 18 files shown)


src/Microsoft.Azure.CosmosRepository/Batching/InMemoryBatchBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+    private void EnsureCapacity()
+    {
+        if (_operations.Count >= BatchConstants.MaxBatchSize)
+        {
+            throw new InvalidOperationException(
+                $"A transactional batch cannot contain more than {BatchConstants.MaxBatchSize} operations.");
+        }
+    }
+
+    private static async ValueTask CreateAsync<TItem>(TItem item, CancellationToken cancellationToken) where TItem : IItem =>
```
IEvangelist, line 115: These helpers construct new `InMemoryRepository<TItem>` instances. Storage is shared, but change notifications are tied to the DI repository instance, so cross-type batch writes bypass the in-memory change feed unlike existing batch APIs and Cosmos. Please route operations through the existing repository or a shared publisher and add a change-feed test.

src/Microsoft.Azure.CosmosRepository/Batching/InMemoryBatchBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+            throw new ArgumentNullException(nameof(id));
+        }
+
+        EnsureCapacity();
+        _operations.Add(cancellationToken => DeleteAsync<TItem>(id, cancellationToken));
+
+        return this;
+    }
+
+    public async ValueTask ExecuteAsync(CancellationToken cancellationToken = default)
```
IEvangelist, line 81: The Cosmos builder rejects a batch when its item types map to different containers, but this in-memory builder tracks no type/container metadata and executes the same invalid batch. Please enforce the same-container invariant in-memory and add a test using two differently configured containers.

src/Microsoft.Azure.CosmosRepository/Batching/InMemoryBatchBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+    public IBatchBuilder ReplaceItem<TItem>(TItem item) where TItem : IItem
+    {
+        if (item is null)
+        {
+            throw new ArgumentNullException(nameof(item));
+        }
+
+        ValidatePartitionKey(item);
+        EnsureCapacity();
+        _operations.Add(cancellationToken => UpsertAsync(item, cancellationToken));
```
IEvangelist, line 35: `ReplaceItem<TItem>` queues `UpsertAsync`, so replacing a missing ID creates the document and succeeds in memory. Cosmos transactional batches return 404 and roll back in this case, which means the in-memory provider can produce false-positive tests. Please implement true replace semantics (existence check plus the repository's NotFound behavior) and cover the missing-item case.

src/Microsoft.Azure.CosmosRepository/Batching/DefaultBatchBuilder.cs
```diff
@@ -0,0 +1,173 @@
...
+    public async ValueTask ExecuteAsync(CancellationToken cancellationToken = default)
+    {
+        if (_operations.Count == 0)
+        {
+            throw new ArgumentException(
+                "Unable to perform batch operation with no items",
+                nameof(_operations));
+        }
+
+        // This call ensures that all the types are valid for the same container
```
IEvangelist, line 47: High / feature blocker: with the default `ContainerPerItemType = false`, all item types share the physical `ContainerId`, but `GetContainerAsync(IReadOnlyList<Type>)` compares each type's logical `ItemConfiguration.ContainerName` (defaulting to the CLR type name). A normal `Foo + Bar` cross-type batch therefore throws "not all configured to use the same container" even though both physically use the same default container. The new acceptance tests mask this by forcing `ContainerPerItemType = true` and assigning both models the same name, and they are all skipped. Fix: when `ContainerPerItemType == false`, validate against the common physical `ContainerId`; only compare per-type names in per-item-container mode. Add a non-skipped default-options test.

src/Microsoft.Azure.CosmosRepository/Batching/DefaultBatchBuilder.cs
```diff
@@ -0,0 +1,173 @@
...
+        Container container = await containerService.GetContainerAsync(_seenTypes.ToList())
+            .ConfigureAwait(false);
+
+        TransactionalBatch batch = container.CreateTransactionalBatch(new PartitionKey(partitionKey));
+
+        foreach (PendingOp operation in _operations)
```
IEvangelist, line 53: Medium: `PendingOp.Item` is typed as `object?`, so these calls bind as `TransactionalBatch.CreateItem<object>` / `ReplaceItem<object>` / `UpsertItem<object>`. The Cosmos SDK then invokes `CosmosSerializer.ToStream<object>`, erasing `TItem`. Since the library publicly permits a custom `CosmosClientOptions.Serializer`, serializers or source-generated contexts that select metadata by the generic `T` can fail or emit the wrong shape. The default Newtonsoft serializer happens to use the runtime type, so current tests pass. Fix: retain typed pending operations/delegates that invoke the SDK with the original generic item type; add a test with a serializer asserting the received `T`.

src/Microsoft.Azure.CosmosRepository/Repositories/IBatchRepository.cs
```diff
@@ -28,6 +28,13 @@ namespace Microsoft.Azure.CosmosRepository;
 public interface IBatchRepository<in TItem>
     where TItem : IItem
 {
+    /// <summary>
+    /// Creates a fluent batch builder for the given partition key.
+    /// </summary>
+    /// <param name="partitionKey">The partition key for the batch.</param>
+    /// <returns>An <see cref="IBatchBuilder"/> for composing a batch.</returns>
+    IBatchBuilder Batch(string partitionKey);
```
IEvangelist, line 36: Compatibility break: adding the abstract `Batch(string)` member to the public `IBatchRepository<T>` (and transitively `IRepository<T>`) breaks any external implementations or decorators of these interfaces. HEAD is only ~14 commits past tag 10.0.2, so this cannot safely ship as a patch/minor. Fix: introduce a separate capability interface or provide a compatible default implementation, otherwise reserve this for a major release.

src/Microsoft.Azure.CosmosRepository/Batching/InMemoryBatchBuilder.cs
IEvangelist, line 81: Test-fidelity risk: this in-memory path runs queued operations sequentially with no rollback, so an earlier operation stays committed when a later one fails, concurrent readers can observe partial state, and a pre-cancelled token still mutates because the underlying in-memory CRUD ignores cancellation. Public `IBatchBuilder` presents these as transactional (the non-atomic remark lives only on the internal class), so this can yield false-positive rollback/cancellation tests. Fix: stage/validate then commit under partition/container synchronization, check cancellation before commit, or explicitly expose and document a non-transactional test-double contract.

marcselman:
Nice work — the fluent `IBatchBuilder` is a much better shape than the per-type `*AsBatchAsync` methods for anything touching more than one type.

A scope question rather than review feedback: is a patch operation in scope for `IBatchBuilder`, or deliberately out?

```csharp
IBatchBuilder PatchItem<TItem>(string id, Action<IPatchOperationBuilder<TItem>> builder) where TItem : IItem;
```

The reason I ask is that Cosmos caps a single-document patch at 10 operations, and `UpdateAsync(id, builder, ...)` currently sends the whole list straight through, so an 11th operation fails with a bare 400 (filed separately as #547). The fix wants a `TransactionalBatch` — and the plumbing this PR adds (container resolution, partition-key validation, batch execution, the in-memory equivalent) is exactly what that needs. A `PatchItem` on the builder would also let a wide patch compose atomically with other work in the same partition, which nothing in the library can express today.

I'd rather not duplicate or conflict with what you have here, so: if patch is in scope, I'm glad to help with it in this PR. If you'd rather keep this focused and get it merged, that seems entirely reasonable too — I'd then do it as a follow-up on top of your `IBatchBuilder` once this lands, and I'll wait rather than build against a moving target.

@IEvangelist — same question to you on whether batched patch belongs on this interface at all.

